PR #70, opened 2016-12-16
Solving problem with migrating data to postgres

gomex:
Solving issue #60

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,6 @@ COPY .env /code/.env
 COPY manage.py /code/manage.py
 COPY jarbas /code/jarbas
 WORKDIR /code
-#RUN python manage.py migrate
-#RUN python manage.py ceapdatasets
 RUN echo "America/Sao_Paulo" > /etc/timezone && dpkg-reconfigure -f noninteractive tzdata
 VOLUME /code/staticfiles
 CMD ["gunicorn", "jarbas.wsgi:application", "--reload", "--bind", "0.0.0.0:8001", "--workers", "4"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ COPY .env /code/.env
 COPY manage.py /code/manage.py
 COPY jarbas /code/jarbas
 WORKDIR /code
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -7,8 +7,8 @@ COPY .env /code/.env
 COPY manage.py /code/manage.py
 COPY jarbas /code/jarbas
 WORKDIR /code
-RUN python manage.py migrate
-RUN python manage.py ceapdatasets
+#RUN python manage.py migrate
+#RUN python manage.py ceapdatasets
```
cuducos: If that's the way, can we delete these lines instead of commenting them? I feel like leaving them commented may cause some confusion.

Dockerfile-elm
```diff
@@ -3,7 +3,9 @@ COPY elm-package.json elm-package.json
 COPY package.json package.json
 COPY gulpfile.js gulpfile.js
 COPY jarbas jarbas
+COPY contrib/copyassets.sh copyassets.sh
+RUN chmod +x copyassets.sh
 RUN npm install
 VOLUME /assets
 RUN npm run assets
-CMD ["cp", "-r", "jarbas/frontend/static/*", "/assets"]
```
cuducos: Just out of curiosity: any way to explain to a newbie why this doesn't work properly, but running the same thing from a Shell script works?

Makefile
```diff
@@ -1,5 +1,7 @@
 run.jarbas:
-	docker-compose up -d
```
cuducos: Can we update the `README.md` with these changes too?

Dockerfile
gomex: I agree!

Dockerfile-elm
gomex: I really don't know why this didn't work :( Shame! but I copied to script to avoid that problem.

Makefile
gomex: Absolutely agree! I will do that now.

cuducos:
Once more, many thanks, Gomex ; )

PR #71, opened 2016-12-16
Format prices in the UI

### LINKED ISSUE #67: Issue with 'Document value' presentation
By going through http://jarbas.datasciencebr.com/#/party/PMDB, I noticed the values of 'Document value' are presented like ##.###. As we use English, it should be like ##.###,##: 1) use '.' as thousands separator; 2) ',' for decimal separators; 3) use only 2 decimal places. By spotting` Document value "60.000" ` I interpret like a spending of R$ 6000,00 instead of R$ 60,00 like is on the receipt.

cuducos:
Fix #67


Irio:
Been waiting for this moment. 🎉

Issue #62, opened 2016-12-12
Cleanup: remove old models, serializers, views, tests etc.


cuducos:
Once #52 is done, we can do some house keeping:

  - [x] Remove old models (`Document`, `Receipt`)
  - [x] Remove old serializers (`DocumentSerializer`, `ReceiptSerializer`)
  - [x] Remove old views (`DocumentViewset`, `ReceiptViewSet`)
  - [x] Remove `OldLoadCommand`
  - [x] Rename suppliers to company
    - [x] model
    - [x] views
    - [x] serializers
    - [x] tests
    - [x] urls
    - [x] documentation
    -  [x] fron-end
  - [x] Rename `NewReceipt` object to `Receipt` (at `models.py` and subsequent views)
  - [x] Rename command `loadsuppliers` to `companies`
  - [x] Use `RetriveAPIView` instead of `Viewset` for companies
  - [x] Remove Django Admin
  - [x] Remove mentions to `loadsuppliers` and `AMAZON_DATASETS_DATE`
    - [x] `README.md`
    - [x]  `.env` and `contrib/.env.sample`
    - [x]  `settings.py`

Issue #61, opened 2016-12-09
New UI/UX


cuducos:
Now Jarbas is changing. From an internal tool for Serenata it will become the public API for the general public. And this impacts UI/UX.

In terms of UX we might change the landing page to offer two ways to start a search, to start browsing  our data, maybe two tabs or something like that (sorry about the `pt-BR`):

![](https://imgur.com/Jl1uwXh.png)

And in terms of UI, @tatianasb is working on it.

This is want we have in mind, but we are open to discuss how to better conduct this changes (the technical part is discussed in #52). I just ask you to be down-to-earth and precise in further comments, suggestions — we have a deadline with our [1296 bosses](https://catarse.me/serenata) which is mid-January ; )

felipekzig:
@cuducos 
How is this redesign going? I'm thrilled to help on the UI piece, although I'm bit new to elmJS. Let me know if this issue is still wanted and if you have any other design in mind and which we should start to move forward.

Thanks,
Felipe

cuducos:
Hi @felipekzig — just to clarify: are you willing to contribute with the graphic design part (mostly visual) or with the coding it into HTML/CSS/Elm whatever?

felipekzig:
Hey @cuducos,
Definitely coding contribution – I could help on the visual, but this is not my strongest skill.

cuducos:
Ok, great then… Probably we'll keep the Elm for the front-end, but I used the [material design library](https://github.com/debois/elm-mdl) and I don't think we need this level of abstraction.

Probably the new version is going for Elm + standard CSS built from SASS (I'm big fan of [Bourbon](http://bourbon.io) family, but no strings attached). Once we have the visual done we can get started. Wanna join us? Drop a line on Telegram ([public group](https://telegram.me/joinchat/AKDWc0BDOqriD1n-mntRBg) ou PVT me, same username).

cuducos:
Ok, here it's, @tatianasb's proposal:

![](http://imgur.com/7Y167GP.png)

fgrehm:
That looks nice! How would the recently introduced "Reimbursements on the same date" fit into this? Would it be just a "replica" of the "Reimbursement details" section?

felipekzig:
Great! This looks amazing. How are you willing to split it up into workable items? And how are you planning to track them?

I thought about the following items:
1. Page overall structure
2. Sections
    2.1. Congressperson
    2.2. Document
    2.3. Company
    2.4. Reimbursement

Thank you

[...]

PR #73, opened 2016-12-19
Add new reimbursements dataset fields to the UI

cuducos:
Partially cover #61…
…and make room for #62!


Irio:
Since it's changing the UI, can we have a print screen?

cuducos:
**Search screen**

Search screen

**New layout for displaying reimbursement data**

New layout for displaying reimbursement data 

**If is there a one (or more) suspicion, it is listed***

If is there a one — or more — suspicion, it is listed

PR #75, opened 2016-12-20
Cuducos add currency link

cuducos:


Irio:
Awesome. This will be super useful for auditors.

PR #76, opened 2016-12-20
Add command to fetch receipts

cuducos:
Takes ages, must be rewrite using lauching multiple `Reimbursement.get_receipt_url()` in parallel…

… but that's what we've got now.


Irio:
Good enough for now. Thanks.

PR #82, opened 2016-12-22
Allow multiple values in search

cuducos:
Query combining `AND` (for different fields) and `OR` (for different values for the same field). Allows to list more than one `document_id` per API request, for example:


Irio:
Can we allow just the usage of one of the characters (my suggestion is `,`) for separation? Maintaining multiple options may be troublesome with the time (having to chose the preferred for helping beginners, docs...)

cuducos:
Sure we can. But I'd opt for ` ` (blank space). It has t be escaped (as `+`) in the API direct access, but it would be the default users type in Jarbas GUI I guess…

That's why I added both: `,` seems the default for API usage, ` ` the default for GUI usage. No need to over documenting as those are probably users' first options. Yet, if users type `, ` (comma plus space) it'd also work as it is ; )

(Gonna fix the conflicts meanwhile…)

Issue #59, opened 2016-12-08
Settings instructions is relevant for Docker users too


cuducos:
The [Settings](https://github.com/datasciencebr/jarbas#settings) is relevant to Docker users too, whe should reorganize that in the `README.md`. Thanks @danizavtz

PR #85, opened 2016-12-22
Deploy Jarbas using Ansible & Docker

gomex:


gomex:
Better documentation related of #59 issue that is already closed.

cuducos:
I have the feeling that it adds a lot of duplicated content. Once we change something we have to update in two different places, it's easy to forget and ending up with outdated documentation.

Maybe we could make a settings section before _Docker installation_ and _Local installation_ sections… what do you think?

cuducos:
Renaming this PR as it looks like as an enhancement on deployment now ; )

.travis.yml
```diff
@@ -1,13 +1,17 @@
+---
 sudo: false
 dist: trusty
 
 language: python
 python:
-  - 3.5
+  - 2.7
```
cuducos: Not sure what it happens as a consequence of this change here. Ubuntu Trusty (`dist: trusty` a couple of lines above) has Python 3 as a default. So I'm not sure which version is used with `python` and `pip` (my guess: Python 3), and not sure if a) is there a `python2` or `pip2`, and b) what Travis does when you set a Python version different from the Trusty default.

Saying we have two pairs of binaries, one for Python 2 (e.g. `python2` and `pip2`) and another for Python 3 (e.g. `python3` and `pip3`), I'd say the best shot would be to:

1. Install Ansible _globally_ (`pip2 install ansible`)
2. Create and activate a virtualenv with Python 3 in the `before_script` section (`python3 -m venv .venv && .venv/bin/activate`)
3. And from now on Jarbas will run within the virtualenv
4. Deactivate the virtualenv when Ansible is needed (`deactivate`)

README.md
```diff
@@ -128,6 +128,36 @@ There is also a [tapioca-wrapper](https://github.com/vintasoftware/tapioca-wrapp
 
 ### Using Docker
```
cuducos: Would you mind rolling back the changes on `README.md`? (see [my previous comment](https://github.com/datasciencebr/jarbas/pull/85#issuecomment-268895808))

docker-compose.yml
```diff
@@ -12,6 +12,11 @@ services:
        - DATABASE_URL=postgres://jarbas:mysecretpassword@db/jarbas
        - CACHE_BACKEND=django.core.cache.backends.memcached.MemcachedCache
        - CACHE_LOCATION=cache:11211
+       - DEBUG=False
```
cuducos: If we move all envvars here, I can think of two consequences:

* A good one: we can leave the `.env` file just for non-Docker users, which is good, reduces the chance of having different versions values here (`docker-compose.yml`) and there (`.env`).
* A maybe not so good one: it kind of breaks the **12 factor** idea of [storing config in the environment](https://12factor.net/config) (not in the project files, repo…)

Is there a way to make Docker read those info from the environment so we avoid keeping them hardcoded here? Or maybe we can go for a `docker-compose.yml.example` that people rename and config by themselves…

roles/digitalocean/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
```
cuducos: Should we customize this file? I couldn't grasp it's idea. Is there a conflict between the licenses (Jarbas uses MIT)?

roles/digitalocean/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
...
+    ip: "{{ my_droplet1.droplet.ip_address }}"
+    api_token: "{{ do_token }}"
+
+- debug: msg="ID is {{ my_droplet1.droplet.id }}"
+  when: do_state == "present"
+
+- debug: msg="IP is {{ my_droplet1.droplet.ip_address }}"
+  when: do_state == "present"
+
+- name: ensure droplet two exists or remove it
```
cuducos: Not sure if it's a problem: we actually don't have Digital Ocean account. We're given access to a droplet (basically they add our keys so we login without password via SSH). What I mean is that we might not be able to spin new droplets and discard others… if it's crucial I can reach them and ask about it, ok?

gomex:
Sorry man! I forgot to create a new branch and this PR was compromised. I will close this and open a new one soon.

PR #91, opened 2016-12-26
Make same day links opens in new tab/window

cuducos:


Irio:
Screenshot? :)

cuducos:
The hasn't changed, still the same as in #86  (i.e. as in production), that's why I haven't added a screenshot. The only change is that clicking the link opens the reimbursement in a new tab.

PR #96, opened 2016-12-29
Concurrent version for receipts command

### LINKED ISSUE #77: Make receipts command faster
Multithreading and handling multiple requests in parallel are the key (it's the `Reimbursement.get_receipt_url()` from #76 that handles the HTTP requests)

cuducos:
Fix #77 

Some additional comments:

* In spite of what I said in #77`asyncio` is not HTTP friendly; thus instead of inviting another external dependency (`aiohttp`) in, I opted for multithreading.
* While testing this script apparently I was blocked after ~400 requests; that's why the script got is a bit more complex than it was before — I implemented a kind of batch size and pause to avoid blocking. These limits are configurable and you can play by yourselves (`python manage.py receipts --help` to get started).

And finally a question:

I hardcoded the `max_workers`, i.e the number of threads. Does it worth it to add an argument (as I did for batch size and pause) or just leave it as it is?


cuducos:
Quick note: despite the red cross here, tests for this PR are not failing (all Python tests pass and Elm tests timeout):

```console
The job exceeded the maximum time limit for jobs, and has been terminated.
```

This is the issue that I fixed at #97.

Just to be sure, once #97 is merged I bring the changes to this PR.

PR #97, opened 2016-12-29
Make Elm tests (way) faster in Travis

cuducos:
Based on https://github.com/elm-lang/elm-compiler/issues/1473

**TL;DR**

Travis CI says to the Elm compiler we have 36 cores when we actually have 2. That's why compiling Elm takes ages, causing _timeouts_ and kind of unreliable Travis CI results (as tests haven't actually failed). The fix came from the link mentioned above.


cuducos:
Ready to _code review_ ; )

.travis.yml
```diff
@@ -14,21 +18,33 @@ services:
 env:
   - DATABASE_URL="postgres://postgres@localhost/jarbas"
 
-install:
-  - psql -U postgres -c 'create database "jarbas";'
+before_install:
+  - |
+    if [ ! -d sysconfcpus/bin ];
```
Irio: Will be something pretty hard to understand why this is here in the future. Luckily you already documented in the commit.

cuducos: As it's a _known_ issue within Elm community, I hope that two things are gonna follow:

* This will be so common for Elm users that we'll get used to it
* @travis-ci is going to fix it or offer a better solution (as they did with `travis_wait` command, for example)

Let's wait and see…

PR #99, opened 2017-01-02
Add panel with reimbursements made in the same month and related to the same subquota

cuducos:
If tests fail, probably it's related to #97 — once it's merged to `master` I can re-run tests here if needed.


Irio:
Awesome.

PR #102, opened 2017-01-04
Avoid stopping bulk update on HTTP error (receipts command)

cuducos:
The `receipts` command would break if we get a HTTP error. Not anymore ; )

### DIFF
--- a/jarbas/core/models.py
+++ b/jarbas/core/models.py
@@ -1,6 +1,7 @@
 from django.contrib.postgres.fields import JSONField
 from django.db import models
 from requests import head
+from requests.exceptions import ConnectionError
 
 from jarbas.core.querysets import SameDayQuerySet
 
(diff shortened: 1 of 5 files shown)


jarbas/core/models.py
```diff
@@ -11,9 +12,10 @@ def __init__(self, year, applicant_id, document_id):
         self.year = year
         self.applicant_id = applicant_id
         self.document_id = document_id
-        self.url = self.get_url()
+        self.exists_cache = None
```
Irio: I'm fairly confused about this variable name - variable which may contain an Exception or a Bool value. Can't think about a good name to fit both types, maybe a good reason for refactor?

cuducos: Sure thing! Thanks for the heads up ; ) Gonna send a new commit soon

cuducos:
Way better now ; )

Issue #101, opened 2017-01-04
RFC: Ability to filter reimbursements by a range of 'issue_date's


fgrehm:
The other day I was playing with the flight tickets dataset and was able to identify tickets for international flights (basically check if the `total_net_value` is > 4k).

With that info we'd be able to identify a period of time when the congressperson was abroad and investigate reimbursements in Brazil that requires the deputy to be physically present (like taxi and meals).

I believe we have more important things to cover before getting to the point of automating the extraction of "period away from Brazil" data points BUT I'm willing to manually check some of those reimbursements to ensure it is worth the trouble to automate this in rosie in the future and for that it would be super helpful if jarbas could provide a search for the following filters in combination:

- issue_date_start
- issue_date_end
- congressperson_id (applicant_id might be enough as well)
- subquota_number

Dunno if it would be useful for other use cases so feel free to 👎 this issue 😄

cuducos:
Closed by PR #106

PR #106, opened 2017-01-04
Add date range search

cuducos:
As proposed in #101


jarbas/core/querysets.py
```diff
@@ -35,13 +35,36 @@ def list_distinct(self, field, order_by_field, query=None):
...
+            'year': (1994, 1996),
+            'applicant': (1, ),
+        }
+    """
+    rx = re.compile('[ ,]+')
+    return {k: tuple(rx.split(v)) for k, v in filters.items()}
+
+
+def _rename_key(key):
+    mapping =  dict(
```
Irio: ✂️ a duplicated space here.

cuducos: Nasty duble space 😠 (fixed in 5d74c95)

jarbas/frontend/elm/Documents/Inputs/View.elm
```diff
@@ -10,6 +10,42 @@ import Material.Textfield as Textfield
...
+matchDate : String -> Bool
+matchDate value =
+    if String.isEmpty value then
+        True
+    else
+        let
+            regex : Regex.Regex
+            regex =
+                Regex.regex
+                    "[\\d]{4}-[\\d]{2}-[\\d]{2}"
```
Irio: Is the regex implemented by different from the default? Otherwise `\d{4}\-\d{2}\-\d{2}` might do the trick (will still allow numbers out of the expected range, but looks good enough for me).

cuducos: Actually you have to escape the slashes, so you suggestion would transform in `\\d{4}\\-\\d{2}\\-\\d{2}` which I found kind of confusing to read — so I opted for the brackets. But as far as I can remember both worked when I was coding it.

rogeriochaves: It does look weird having to escape the slashes, bikeshedding a bit, that gave me the idea: what if we had a lib for building type-safe regexes? Turns out that there is already one! http://package.elm-lang.org/packages/ydschneider/regex-builder/1.0.0/RegexBuilder

Probably overkill and maybe not very friendly for other developers as one of the advantages of regexes is that they are mostly universal across languages.

cuducos: I'm not so worried right now. We have plans to change this UI/UX completely (#61) and probably use a more human-friendly date selector…

Anyway very interesting package this RegexBuilder — totally gonna mess around with it!

PR #115, opened 2017-01-12
Elm and Django need to run in developer mode

gomex:


Dockerfile-elm
```diff
@@ -1,13 +1,10 @@
 FROM node:6.9.1
+RUN npm install -g yarn gulp
```
cuducos: Gulp is installed in line 8 (`yarn install`) already, as it is listes in `packages.json`.

docker-compose.yml
```diff
@@ -6,12 +6,13 @@ services:
...
     volumes:
        - "staticfiles:/code/staticfiles"
-       - "assets:/code/jarbas/frontend/static"
+       - "./:/code"
     environment:
        - SECRET_KEY=lets-rock
        - DATABASE_URL=postgres://jarbas:mysecretpassword@db/jarbas
        - CACHE_BACKEND=django.core.cache.backends.memcached.MemcachedCache
        - CACHE_LOCATION=cache:11211
+       - DEBUG=yes
```
cuducos: Isn't it suppose to be `DEBUG=True`?

PR #116, opened 2017-01-12
Remove the current reimbursement from the same subquota panel

cuducos:
Example (with current reimbursement repeated in the _same subquota_ panel): https://jarbas.datasciencebr.com/#/documentId/5893871

### DIFF
--- a/jarbas/frontend/elm/Documents/RelatedTable/Model.elm
+++ b/jarbas/frontend/elm/Documents/RelatedTable/Model.elm
@@ -25,6 +25,7 @@ type alias Results =
 
 type alias Model =
     { results : Results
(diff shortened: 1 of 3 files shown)


Irio:
I've seen this happening (not sure if was a single time or with every reimbursement). Was appearing all the times?

cuducos:
Yep. It was a bug.

Not for the same day panel though (this panel uses an specific endpoint and the current reimbursement is excluded from the results in the backend).

However the same subquota panel was implemented using the filtering options of the existing API which had no exclude option. So that's what this PR fixes: front-end validation of the results to avoid showing the current reimbursement in the panel.

PR #130, opened 2017-01-15
Elm refactor: performance and readability

cuducos:
Despite minor enhancements this PR is focused on enhancing performance and readability by:

* Using `Array` (instead os `List`) to store data that needs to be accessed by index (the main reimbursement list, and reimbursement summaries in the _same day_ and _same subquota_ panels)
* Using `Dict String String` (instead of `List ( String, String )`) to check and manipulate URL parameters (for example, whenever we need to check the hash from URL for a `page` parameter)

In the _minor enhancements_ list I'd mention: changing some data types, renaming things to match the current abstraction of CEAP, compiling settings, organizing imports, some clean up etc.


rogeriochaves:
LGTM, naming it Reimbursement instead of Document is indeed much better!

Issue #124, opened 2017-01-14
Store company activities as JSONField


cuducos:
Nowadays each `Company` stores its main and secondary activities in a separate model (`Activity`) using a `ManyToManyField`.

I don't think it's necessary as we won't query by a company activity (at least not that frequently). Also this design makes the `python manage.py companies <companies.xz>` extremely slow because we cannot use `bulk_create` with relational tables (or can we?).

Guilhermeslucas:
So, you're thinking to replace the activity model with a field on the companies model as JSON, with all the primary and secondary activities?

cuducos:
Yes, that's it.

Basically we can (just drafting here, probably I've forgotten some steps, such as checking if Django REST Framework serializers are consistent with the new strategy etc.):

1. Edit `models.py` removing `Activity` and setting up `main_activity` and `secondary_activity` to `JSONField`
2. Create a migration ; )
3. Adapt the serializer inside `companies.py` to parse `main_activity` and `secondary_activity` as something like a list of `dict`s (eg. `[ {"code": "42", "description": "foo"}, {"code: "31415", "description": "bar"} ]`)
4. Use `bulk_create` (as in `reimbursements.xz`) to speed up the import of companies data

cuducos:
An extra would be to manage data in the migrations, so we don't have to loose data and reimport it when moving along the migrations history…

Guilhermeslucas:
Ok @cuducos. I'll give it a try. Now, to check these informations we check the /api/company/<cnpj>/ adress, right?

cuducos:
Yep. To check this info you get to `/api/company/00000000000000/` where this tons of zeros is a CNPJ ; )

Guilhermeslucas:
I changed the model an the Django REST Framework is fully working. But when I make a search on the index page of Jarbas, it doesn't show the campanies detail, as in the pictures above. Any idea on how to fix it?

Working REST Framework:
![captura de tela de 2017-01-18 00 58 38](https://cloud.githubusercontent.com/assets/9682164/22049278/6f904f2c-dd19-11e6-9c4e-f0f3cbf37e99.png)

Search not displaying companies information:
![captura de tela de 2017-01-18 00 58 53](https://cloud.githubusercontent.com/assets/9682164/22049279/76008e6c-dd19-11e6-969b-78d54d5b05dc.png)

Guilhermeslucas:
[...]

PR #131, opened 2017-01-17
fixing text. Now the api prints the correct JSON

Guilhermeslucas:
This is a really simple change I've made while working on the #124 issue. On the skip list, activity was written wrong, so the API was making a not full formated JSON.


cuducos:
Yay! Many thanks 🎉

PR #133, opened 2017-01-17
Minor enhancements in Elm & nginx for Docker 

cuducos:
* c819b4e : Instead of compiling `app.js` to `/code/jarbas/frontend/static` and wait for the `collectstatic` command copy it to `/code/staticfiles`, now Elm container compiles `app.js` directly into  `/code/staticfiles` for nginx to serve it.
*  9a0749c : Rename the working directory in the nginx container to match the path in other containers


cuducos:
cc @gomex — does it make sense?

gomex:
Make sense :)

PR #129, opened 2017-01-15
Introducing docker prod environment 

gomex:
We still need to put secrets inside docker-compose-prod and somethings, but let's try first

PR #137, opened 2017-01-18
Setup a new way to compose containers

pedrommone:


docker-compose.yml
```diff
@@ -6,7 +6,6 @@ services:
       context: .
       dockerfile: Dockerfile
     volumes:
-      - "./:/code"
```
gomex: We need this to keep developing using yarn watch

docker-develop.yml
```diff
@@ -0,0 +1,12 @@
...
+
+  jarbas:
+    volumes:
+      - "./:/code"
+
+  elm:
+    volumes:
+      - "./:/code"
+      - "/code/node_modules"
+    command: [yarn, watch]
```
gomex: The ideia is keep development in default docker-compose file.

cuducos:
@gomex and @petriuslima — Pedro's PR fixes some bug we had in the current Docker environment (compiled files from Elm container was missing for nginx container). However it changes a bit the way things work. For production the `docker-compose.yml` is the base, and you load the `docker-develop.yml` that adds certain configuration that should make it a proper development environment. Does that make sense, can we merge it?

gomex:
@cuducos IMHO development opinion should be easy to use. In that case docker-compose.yml should be for dev.

About that Prod vs Dev env I already sent this PR #129

cuducos:
Ok… we have three scenarios here:

* `master` was supposed to be focused on dev, but it was broken
* #129 introduced Docker for production, but as it was based on `master` development was still broken
* Now we have a third approach (this PR) that attempts to fix both

I do agree that development should be easy to use, and that production should be made via Docker too — this is our target. I'm just not experienced enough to fix all Docker issues and to ponder on these three scenarios.

Today I'll try #129 for _dev_ and _prod_, and this PR #137 for _dev_ and _prod_  on a fresh machine and gonna update you all.

cuducos:
Hi all,

I tested this PR 3 times for _dev_ 3 times for _prod_, on a fresh VM. No good news — and I have no idea what we did yesterday that make it work.

**Production**

```console
django.db.utils.OperationalError: could not connect to server: Connection refused
	Is the server running on host "db" (172.20.0.2) and accepting
	TCP/IP connections on port 5432?
```

But the PostgreSQL container was up:

```console
$ docker-compose ps
     Name                    Command               State                Ports
-----------------------------------------------------------------------------------------
jarbas_cache_1    docker-entrypoint.sh memcached   Up       11211/tcp
jarbas_db_1       /docker-entrypoint.sh postgres   Up       5432/tcp
jarbas_elm_1      /bin/sh -c yarn assets           Exit 0
jarbas_jarbas_1   gunicorn jarbas.wsgi:appli ...   Up
jarbas_nginx_1    nginx -g daemon off;             Up       443/tcp, 0.0.0.0:8000->80/tcp
```

This test routine was as follows:

```console
git clone https://github.com/datasciencebr/jarbas.git && \
cd jarbas && \
git pull https://github.com/pedrommone/jarbas.git cuducos-elm-to-nginx && \
sed 's/127.0.0.1/138.197.13.210/g' contrib/.env.sample > .env && \
docker-compose up -d --build && \
docker-compose run --rm jarbas python manage.py migrate && \
docker-compose run --rm jarbas python manage.py ceapdatasets && \
docker-compose run --rm jarbas python manage.py collectstatic --no-input
```

I've seen this database situation a couple of times in Docker and I know that usually re-running stuff fixes it. However in my three tests I ended up there.

**Development**

The three times I trie to build a development environment the database issue was gone. But it failed to serve the static folder:

```console
$ curl -I localhost:8000/static/app.js
HTTP/1.1 404 Not Found
Server: nginx/1.11.5
Date: Thu, 19 Jan 2017 15:02:15 GMT
Content-Type: text/html
Content-Length: 169
Connection: keep-alive
```

This test routine was as follow:

```console
git clone https://github.com/datasciencebr/jarbas.git && \
cd jarbas && \
git pull https://github.com/pedrommone/jarbas.git cuducos-elm-to-nginx && \
sed 's/DEBUG=False/DEBUG=True/g' contrib/.env.sample > .env && \
sed 's/127.0.0.1/138.197.13.210/g' .env > .env && \
docker-compose -f docker-compose.yml -f docker-develop.yml up -d --build && \
docker-compose run --rm jarbas python manage.py migrate && \
docker-compose run --rm jarbas python manage.py ceapdatasets && \
docker-compose run --rm jarbas python manage.py collectstatic --no-input
```

cuducos:
Closing by a) inactivity and to b) focus on the active Docker-related PR #163

PR #139, opened 2017-01-20
[WIP] Changing Activities to JSON Field

Guilhermeslucas:
As @cuducos said, I'm opening this PR so you can give me a feedback on what I've done so far. My main concern is with the tests. I commented some tests, that treat activities as a Model. If this is ok, I'll work on bulk create.


Guilhermeslucas:
I added the bulk_create too. But, again, need some help to rewrite test, if possible.

jarbas/api/tests/test_company_view.py
```diff
@@ -11,9 +11,7 @@
 class TestApi(TestCase):
 
     def setUp(self):
-        activity = Activity.objects.create(**sample_activity_data)
```
cuducos: Maybe we need to cleanup the imports too (line 7), don't we?

jarbas/core/management/commands/companies.py
```diff
@@ -10,6 +10,10 @@
 
 class Command(LoadCommand):
     help = 'Load Serenata de Amor companies dataset into the database'
+    
+    def add_arguments(self, parser):
+        super().add_arguments(parser)
+        parser.add_argument('--batch-size', '-b', dest='batch_size', type=int, default=10000, help='Number of documents to be created at a time')
```
cuducos: Can we be a bit more friendly towards PEP8? We don't have to stop strictly at 79, but IMHO this line is too long…

jarbas/core/management/commands/companies.py
```diff
@@ -18,50 +22,56 @@ def handle(self, *args, **options):
 
         if options.get('drop', False):
             self.drop_all(Company)
-            self.drop_all(Activity)
+            #self.drop_all(Activity)
```
cuducos: We can actually delete this line, can't we?

jarbas/core/management/commands/companies.py
```diff
@@ -18,50 +22,56 @@ def handle(self, *args, **options):
...
+                    filtered['secondary_activity'] = [secondary]
+                obj = Company(**self.serialize(filtered))
+                yield obj
+                 
+                #obj.save()
+                #
+                #self.count += 1
+                #self.print_count(Company, count=self.count)
 
     def save_activities(self, row):
```
cuducos: If `Activity` is not a model anymore, I'd add the logic to _parse_ the data from teh CSV into a dict to the `serialize` method, not here… this method is called _save_ something, but it doesn't save anything, but actually serialize activities ; )

jarbas/core/tests/test_companies_command.py
```diff
@@ -88,20 +71,20 @@ class TestConventionMethods(TestCommand):
...
     def test_handler_without_options(self, print_count, save_companies, drop_all, print_):
         print_count.return_value = 0
-        self.command.handle(dataset='companies.xz')
+        self.command.handle(dataset='companies.xz', batch_size=42)
         print_.assert_called_with('Starting with 0 companies')
         self.assertEqual(1, save_companies.call_count)
         self.assertEqual(1, print_count.call_count)
         self.assertEqual('companies.xz', self.command.path)
-        drop_all.assert_not_called()
+        #drop_all.assert_not_called()
```
cuducos: This (uncommented) is still meaningful here, isn't?

jarbas/core/management/commands/companies.py
Guilhermeslucas: Done! Much better now!

jarbas/core/management/commands/companies.py
Guilhermeslucas: Done!

Guilhermeslucas:
Many thanks, @cuducos. I'll make the requested changes ASAP. About the tests, I'll give it a try, on these next days.

Guilhermeslucas:
About the ```save_activities``` method, what if I change the name for ```return_activities``` ? Makes more sense and keeps the code clean.

Guilhermeslucas:
The only problem I'm having right now is with ```companies.xz``` on ```test_companies_command```. The test routine says that it doesn't exists. Where is this file?

cuducos:
> About the `save_activities` method, what if I change the name for `return_activities` ? Makes more sense and keeps the code clean.

Personally I don't agree because we already have a `serialize` method that takes a row from the CSV and returns a serialized company — and the companies activity should be part of this process of serializing each row.

As serializing the activities is part of this logic I think that this method should be within the `serialize` method. Either extracting it logic to a external method (`serialize_activities`, or `_activities_to_dict` for example) and calling it within the `serialize` method, or including its logic within `serialize` method.

> The only problem I'm having right now is with `companies.xz` on `test_companies_command`. The test routine says that it doesn't exists. Where is this file?

The file is not present in the repo. But the main point here is that our tests are not suppose to test the hability of `lzma.open` to actually open this file. We trust `lzma.open` works pretty well so we [mock it](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/tests/test_companies_command.py#L64) during tests ; )

Guilhermeslucas:
Ok! I'll put it into the serializer. Do you have a guess why the test routine says that companies.xz do not exists? It was working before!

Guilhermeslucas:
What do you think avout this last commit? We have no more external routine. If you approve this, i'll get rid of comments and the ```return_activities``` routine. Many thanks!

cuducos:
Hi @guilhermeslucas. I still believe the code is not neat. You moved the logic of serializing companies activity to a method that is responsible for generating companies. What I suggested earlier was to move [the logic of serializing company activities](https://github.com/Guilhermeslucas/jarbas/blob/2cfae2bbaccac13fe0d20aba253549e7eae72364/jarbas/core/management/commands/companies.py#L49-L66) to the [method that serialize companies](https://github.com/Guilhermeslucas/jarbas/blob/2cfae2bbaccac13fe0d20aba253549e7eae72364/jarbas/core/management/commands/companies.py#L87-L98).

Guilhermeslucas:
Sorry about the commit, I got distracted. This last commit os trying to fix the problem. But I see two problems:
- including secondary keys using the for loop on line 43 of ```companies.py``` is not efficient.
- I am inserting the main activity outside of serialize function, cause I didn't found a way out.

Thanks for the reviews, @cuducos. I'd appreciate any suggestions on the last commit cause I think I'm getting close to fix the problem.

jarbas/core/management/commands/companies.py
```diff
@@ -38,52 +38,41 @@ def companies(self):
...
-
-        return data, secondaries
-
+        
+        if len(secondaries) == 0:
+            row['secondary_activity'] = []
+        else:
+            row['secondary_activity'] = [secondaries]
+
+        for num in range(1,100):
```
cuducos: You can do this cleanup:

* with `del(row['secondary_activity_{}_code'.format(num)])` and `del(row['secondary_activity_{}'.format(num)])
* and within the loop of line 60 to avoid looping again ; )

jarbas/core/management/commands/companies.py
```diff
@@ -38,52 +38,41 @@ def companies(self):
...
         for num in range(1, 100):
             code = row.get('secondary_activity_{}_code'.format(num))
             description = row.get('secondary_activity_{}'.format(num))
             if code and description:
                 secondaries[code] = description
-
-        return data, secondaries
-
+        
+        if len(secondaries) == 0:
```
cuducos: `if secondaries` does the job, but I'm not sure `[secondaries]` is what we're aiming for. This:

```python
# a list with ONE dict holding two activities
[
    {
        '1': 'Firts activity',
        '2': 'Second activity'
    }
]
```

Is different than: 

```python
# a list with TWO dicts holding one activity each
[
    {
        '1': 'Firts activity'
    },
    {
        '2': 'Second activity'
    }
]  
```

And yet, what our front-end is expecting is slightly different:

```python
#a list with TWO dicts holding one activity each, with 'code' and 'description' as keys
[
    {
        'code': 1,
        'description': 'First activity'
    },
    {
        'code': 2,
        'description': 'Second activity'
    }
]
```

jarbas/core/management/commands/companies.py
```diff
@@ -38,52 +38,41 @@ def companies(self):
...
+            row['secondary_activity'] = []
+        else:
+            row['secondary_activity'] = [secondaries]
+
+        for num in range(1,100):
+            row.pop('secondary_activity_{}_code'.format(num), None)
+            row.pop('secondary_activity_{}'.format(num), None)
+
```
cuducos: This method should return `row`, shouldn't it?

Guilhermeslucas:
sorry about the shitty code I'm commiting. I made the changes, I think is better now. Is it ok?

Guilhermeslucas:
@cuducos do you know how can I pass one argument to the function inside filter? I want to pass a list with the Company model fields, to avoid looping everytime.

Guilhermeslucas:
Better now!

cuducos:
> Better now!

Well… indeed 🎊 Congrats and many thanks.

I did some minor code style edits, small refactors and fixed tests in [a PR to your branch](https://github.com/Guilhermeslucas/jarbas/pull/1). If you're OK with that changes, I can merge this PR once you merge my suggestions ; )

Guilhermeslucas:
Just merged your suggestions, @cuducos. I'll take a look at the video about mock testing you did, It's really insteresting!

Guilhermeslucas:
Just be sure I made it the right way, I'm a noob, as you saw! hehe

cuducos:
> Just be sure I made it the right way, cause it looks like I didn't. I'm a noob, as you saw!

You're not a noob. Maybe you was one of these days. But this PR proves you're not ; )

> Btw, is it still working over there?

Not sure yet. I'm rewriting the migrations right now. The way you did it forced the user to reimport data after migrating forward or backward (data was lost in both direction). I'm manually writing migrations so data is converted from the relational way to JSON, and the other way around, when users move along migrations history.

Probably I'll merge this PR to to that branch before `master` — and I'll send you a new PR to fix anything that's not working, is that OK? But if you want, feel free to add more commits to this branch. If you share the error message I can help you spot what might be wrong ; )

Guilhermeslucas:
That's ok, @cuducos. I'll post the error message here and try to fix it. Many thanks.

captura de tela de 2017-01-26 14 46 09

cuducos:
It's really bizarre. The full dataset works, just the sample ends up in this error. I'm gonna try to regenerate the sample and replace the file in the repo…

Guilhermeslucas:
yeah, this is really strange! I'm searching for the bug, but it looks like its really the sample that has something wrong.

cuducos:
> yeah, this is really strange! I'm searching for the bug, but it looks like its really the sample that has something wrong

Got it. It was my fault — fixed with 13931c6 🤕

Guilhermeslucas:
Awesome, @cuducos .

PR #147, opened 2017-02-03
[WIP] Optimize data Import (Fix #125)

Guilhermeslucas:
Hi all! I was working to fix #125 . I want some feedback about the work, specially two things:
- [This part of the code](https://github.com/Guilhermeslucas/jarbas/blob/master/jarbas/core/management/commands/reimbursements.py#L18-L20) adds the irregularities path for the reimbursements command. But, in my opinion, it's weird cause, for ```reimbursements_path``` there is no need to use ```--reimbursements_path``` and, in this case, we need to use ```--irregularaties-path```.

- The identation on [this part of the code](https://github.com/Guilhermeslucas/jarbas/blob/master/jarbas/core/management/commands/reimbursements.py#L38-L42) is horrible. I want some suggestions on how to make it better.

Please, tell me everything you think I have to change!

PR #148, opened 2017-02-09
[WIP] Optimize data imports (Fix #125)

### LINKED ISSUE #125: Optimize data import
Nowadays first we import all reimbursements (`python manage.py reimbursements …`) and later we update these reimbursements with irregularities/suspicions (`python manage.py irregularities …`).

If we bring in `pandas` we could merge both datasets and import both datasets together in one command.

An extra would be that we could have the script drafted in [this markdown](https://github.com/datasciencebr/jarbas/blob/1939c49aa667e7984e8a53bf5fbfa3718e1d0811/contrib/sample-data/README.md) file as an actual `.py` script (since `pandas` would be added to the `requirements.txt`).

Guilhermeslucas:
Trying to fix #125, importing reimbursements and irregularities at once. Just fixed the problems shown in [pull request #147 ](https://github.com/datasciencebr/jarbas/pull/147). Give me more feedback so I can start working on tests and so on. Thanks a lot!


jarbas/core/management/commands/reimbursements.py
```diff
@@ -30,21 +34,47 @@ def handle(self, *args, **options):
...
+            yield Reimbursement(**self.serialize(row.to_dict()))
 
     def serialize(self, reimbursement):
         """Read the dict generated by DictReader and fix content types"""
-
-        missing = ('probability', 'suspicions')
-        for key in missing:
-            reimbursement[key] = None
+        
+        if 'probability' in reimbursement:
```
cuducos: I would split this logic here with a class method `to_float` (as we have `to_bool`, it looks like more coherent IMHO). This method could be fail proof returning `None` whenever there's no way to parse a value to `float`:

```python
    @staticmethod
    def to_float(value):
        try:
            return float(value)
        except TypeError:
            return None
```

Thus we can replace this `if`/`else` by

```python
    reimbursement['probability'] = self.to_float(reimbursement.get('probability'))
````

File: requirements.txt [hunk 2]
```diff
@@ -6,6 +6,7 @@ django-cors-middleware==1.3.1
 django-test-without-migrations==0.4
 djangorestframework==3.5.3
 gunicorn==19.6.0
+pandas==0.19.2
```
Guilhermeslucas: The problem is that I really have no idea what was happening. I'm using anaconda to run it now. Do you think we should add some note on README explaining that panda's issues can be solved with anaconda?

jarbas/core/management/commands/reimbursements.py
```diff
@@ -30,21 +34,47 @@ def handle(self, *args, **options):
...
+        suspicions = {}
+        for key in suspicions_keys:
+            if key in reimbursement:
+                suspicions[key] = self.bool(reimbursement[key])
+                del reimbursement[key]
+        
+        if not suspicions:
+            suspicions = None
+
+        reimbursement['suspicions'] = suspicions
```
Guilhermeslucas: Awesome tip!

requirements.txt
cuducos: Hum… if you're using Anaconda we have to change the documentation. But… before we reach this point: I don't think it's a good idea to proceed with this issue and PR. The overhead of requiring to Anaconda is too big to save a couple of minutes in the seed process.

Guilhermeslucas: I'm using anaconda cause it was the only way I found through the problem. But I really don't think it's always necessary. But, if you really think it's better to close this PR, that's ok for me. :)

cuducos: The problem is that `pip install -r requirements.txt` would fail. And requiring a Anaconda is not common for most webdevs — including the way `conda` conflicts with `virtualenv`…

Guilhermeslucas:
The tests are running ok. Shouldn't it break when running the pip install?
( I really don't know it)

Em 10/02/2017 15:56, "Eduardo Cuducos" <notifications@github.com> escreveu:

cuducos:
> The tests are running ok. Shouldn't it break when running the pip install?
> ( I really don't know it)

Tests passing is a good thing, for sure. But if `pip install pandas` fails and we have it by default probably we'll see issues being opened here. In my macOS it worked pretty fine, but was because I already had `numpy` and a couple of compilers installed. What OS do you use? Can you try to install `numpy` and then `pandas` via `pip`? Also maybe for some Linux users `sudo apt-get install python-pandas` or `yum install python-pandas`.

Guilhermeslucas:
I can't make it work without anaconda, have no idea why. I'm running Ubuntu 14.04. That's really awkward, cause ```python manage.py test``` runs without any problem. I'm trying to run it, from the beginning in a virtual machine!

Guilhermeslucas:
It works on a VM running Ubuntu 14.04. What should we do?

cuducos:
> […] What should we do?

Write tests. It shouldn't pass the test if it's not working ; )

I'm really sorry I'm not able to pair or something at the moment. I'm focused in other parts of the project, and, at the end of the day, this is working (I mean, this PR is a enhancement in terms os speed)… but hopefully you get my point. I don't want push `pandas` to the master branch and see the Issues panel flooded with _It's not working_, _`pip` fails_ etc…

cuducos:
Closing this issue:

With [@RosieDaSerenata](https://twitter.com/rosiedaserenata) Jarbas cannot afford to have its database wiped out each time we update reimbursement and suspicions. So this entire strategy to feed our database should be revised. Until now doing it quickly was a priority but from now on doing it without needing to `--drop-all` is the new black ; )

Probably a strategy with _bulk update_ as in the `irregularities.py` will have a better fit.

PR #163, opened 2017-05-10
[WIP] Make Docker great again 

lipemorais:
This is in progress, I just created it earlier to have earlier feedbacks. 

**What is the purpose of this Pull Request?**
The purpose of this PR is make easier contribute and develop for Jarbas using docker-compose and docker.

**What was done to achieve this purpose?**
Initially I fixed elm dockerfile  to make docker work again for development with Jarbas. I have below a list of things to be done to make it ideal.

**How to test if it really works?**
`make run.devel`

**Who can help reviewing it?**
@gomex @Irio @anaschwendler @jtemporal @cuducos 

**What is out of scope**
Is out of scope make this docker-compose ready for production environment, I'm planing to work on it in a next PR.

**TODO**

- [x] Fixes static files in rest framework front end
- [x] Create a jarbas-frontend image in docker hub
- [x] Update Jarbas image in docker hub
- [x] Improve documentation for Docker images
- [x] Update docker-compose.yml to use new images

PR #168, opened 2017-05-17
Update docs (after Docker)

cuducos:
After merging #163 I felt like reorganizing the documentation a bit.

Basically moved stuff around to keep Docker instructions altogether. Also gave more importance to the documentation os the settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Jarbas is in charge of making data from [CEAP](https://github.com/datasciencebr/
 1. [Installing](#installing)
     1. [Using Docker](#using-docker)
     1. [Local install](#local-install)
+    1. [Settings](#settings)
 
 ## JSON API endpoints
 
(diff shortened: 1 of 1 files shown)


lipemorais:
It's Awesome!

Do you think that would be good have some instructions in docker hub explaining a little more about each Jarbas image?

cuducos:
I haven't thought about it… any suggestions?

cuducos:
I haven't thought about it… any suggestions?

lipemorais:
May this is not applicable for our case, so feel free to disagree. 

What I have in my mind is have an explanation based in great Dockerfiles in Docker Hub like [postgres](https://hub.docker.com/_/postgres/) and [python](https://hub.docker.com/_/python/) with just what make sense for us. Like:
-  Add a more meaningful short description for Jarbas images
- Where get help about the images(_I think that this will be specially helpful for new contributors that don't know about Docker_)
- Where Dockerfile that generates this images can be found

jtemporal:
I think as far as the Readme file goes, this PR looks pretty good so I'm merging it...
=)

PR #170, opened 2017-05-25
Load new reimbursements with update_or_create

cuducos:
Our previous command to load reimbursement was fast (thanks `bulk_create`) but we had no control over reimbursements that:

1. were available in previous datasets
2. but — for whatever reason — were deleted

With `bulk_create` the practice was to dump all reimbursements and load a fresh new dataset to our database. We ended up loosing these _deleted_ reimbursements when they were present in the previous dataste but not in the new one.

That is the case when we report a given reimbursement to the Chamber of Deputies and the representative decided to return the money, this reimbursement will soondisappear from the Chamber API (as it's now settled).

With this PR such reimbursements will not disappear from Jarbas — it will remain on Jarbas properly identified with a flag `available_in_latest_dataset`.

This was impossible to be implemented with `bulk_create` so the downside is that loading reimbursements is now way slower than before.

(Thanks @andrealmar for pairing on that)


jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -110,6 +111,7 @@ def _count_results(self, url):
         return len(content.get('results', 0))
 
 
+@freeze_time('1970-01-01 00:00:00')
```
berinhard: One question, do you feel that the tests running slows down when using `freeze_time`? This happened with me, but I've never profiled to check if this was happening.

jarbas/core/tests/test_reimbursement_model.py
```diff
@@ -26,6 +26,17 @@ def test_unique_document_id(self):
         with self.assertRaises(IntegrityError):
             Reimbursement.objects.create(**self.data)
 
+    def test_last_update(self):
+        reimbursement = Reimbursement.objects.create(**self.data)
```
berinhard: **\<advertising\>**
  Ever wondered about using [model_mommy](https://model-mommy.readthedocs.io/en/latest/basic_usage.html)?
**\</advertising\>**

jarbas/core/management/commands/reimbursements.py
```diff
@@ -80,16 +77,16 @@ def serialize(self, reimbursement):
...
-        self.print_count(Reimbursement, count=self.count)
+    def create_or_update(self, reimbursements_as_dicts):
+        for count, reimbursement in enumerate(reimbursements_as_dicts):
+            document_id = reimbursement.get('document_id')
+            if document_id:
+                Reimbursement.objects.update_or_create(
+                    document_id=document_id,
+                    defaults=reimbursement
+                )
+            self.print_count(Reimbursement, count=count + 1)
```
cacarrara: Even if you do not create or update Reimbursement you' re counting it. Is this correct?

jarbas/core/migrations/0024_add_available_in_latest_dataset_field_to_reimbursement.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
cacarrara: Why to separate migration? Is this a project convention?

jarbas/core/tests/test_reimbursement_model.py
cacarrara: I agree with @berinhard ... There's some options for fixture replacement... I would try [mixer](https://github.com/klen/mixer)

jarbas/core/tests/test_reimbursements_command.py
```diff
@@ -89,69 +89,67 @@ def test_serializer(self):
...
-    @patch.object(Reimbursement.objects, 'bulk_create')
+    @patch.object(Reimbursement.objects, 'update_or_create')
     @patch('jarbas.core.management.commands.reimbursements.Command.print_count')
-    def test_bulk_create(self, print_count, bulk_create):
-        self.command.count = 0
-        self.command.bulk_create([1, 2, 3])
-        bulk_create.assert_called_once_with([1, 2, 3])
-        self.assertEqual(1, print_count.call_count)
+    def test_create_or_update(self, print_count, create):
+        reimbursements = (dict(ahoy=42, document_id=1),)
```
cacarrara: If print_count shouldn't be called when reimbursement is not created (without id), you should test this behaviour too. For instance: new test with two reimbursements, only one with id, assert print_count called once.

jarbas/core/migrations/0024_add_available_in_latest_dataset_field_to_reimbursement.py
berinhard: @cacarrara this is how django stores database migrations. This `migrations` directory is always present whenever a new django app is created.

jarbas/core/tests/test_reimbursements_command.py
```diff
@@ -89,69 +89,67 @@ def test_serializer(self):
...
+    @patch('jarbas.core.management.commands.reimbursements.Command.mark_not_updated_reimbursements')
+    def test_handler_with_options(self, mark, print_count, drop_all, create, reimbursements, print_):
+        self.command.handle(dataset='reimbursements.xz', drop=True)
         print_.assert_called_once_with('Starting with 0 reimbursements')
         drop_all.assert_called_once_with(Reimbursement)
-        bulk_create_by.assert_called_once_with(reimbursements, 1)
-
-    @patch('jarbas.core.management.commands.reimbursements.LoadCommand.add_arguments')
-    def test_add_arguments(self, super_add_arguments):
-        parser = MagicMock()
```
jtemporal: `MagicMock` is not being used anymore, could you please remove it from the imports?

jarbas/core/tests/test_reimbursements_command.py
```diff
@@ -89,69 +89,67 @@ def test_serializer(self):
         self.assertEqual(self.command.serialize(input), expected)
 
 
-class TestBulkCreate(TestCommand):
-
-    @patch('jarbas.core.management.commands.reimbursements.Command.bulk_create')
-    def test_bulk_create_by(self, bulk_create):
-        self.command.bulk_create_by(range(0, 10), 3)
-        expected = (
-            call([0, 1, 2]),
```
jtemporal: `call` is not being used anymore, could please remove it from the imports? =)

jarbas/core/migrations/0024_add_available_in_latest_dataset_field_to_reimbursement.py
cacarrara: My doubt is about why do not create these two new fields with only one migration?

jarbas/api/tests/test_reimbursement_view.py
cuducos: I've never noticed it. But this is only the second time ever that I use that, so I might not be the  best person to have a say on that ; )

jarbas/core/tests/test_reimbursement_model.py
cuducos: #171 (despite its number) is awesome ; ) Way better. Thanks @berinhard and @cacarrara!

jarbas/core/management/commands/reimbursements.py
cuducos: Yep… in my sick mind it was supposed to mean that that line of the CSV (`.xz`) was processed (even if not created/update, it was processed). Does that make sense for anybody else?

jarbas/core/migrations/0024_add_available_in_latest_dataset_field_to_reimbursement.py
cuducos: Two reasons:

* Because `last_update` is useful even if we're not using it for flagging stuff as `available_in_latest_dataset` or not
* Also it reflects the _baby steps_ methodology (as one can see in the [commits](https://github.com/datasciencebr/jarbas/pull/170/commits))

jarbas/core/tests/test_reimbursements_command.py
cuducos: > If print_count shouldn't be called when reimbursement is not created

But IMHO it should…

jarbas/core/tests/test_reimbursements_command.py
cuducos: Damn Code Climate… I miss Landscape.io

jarbas/core/tests/test_reimbursements_command.py
cuducos: Damn Code Climate… I miss Landscape.io

jarbas/core/tests/test_reimbursements_command.py
cuducos: Done in 8e48c36

jarbas/core/tests/test_reimbursements_command.py
cuducos: Done in 8e48c36

jarbas/core/management/commands/reimbursements.py
cacarrara: I raised this question because [print_count()](https://github.com/datasciencebr/jarbas/blob/36fc5f70087e6c9aef3fb8dfe5751f6494f18e29/jarbas/core/management/commands/__init__.py#L58) takes only objects created as default. Anyway, both outputs seems relevant. What about two logs? One for created another for processed.

cuducos: But the model, [when `count` is passed, is used only just for `__name__` sake](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/__init__.py#L57-L63)…

cacarrara: Yes. It's only a metter of usability. Since print_count() default is created objects, I suppose the users are used to take that log as "created objects". Since here it's not the same info, but with same output, I got worried about some misunderstanding.

cuducos: I still tend to disagree…

The printed output is simply:

> Current count: 42 reimbursements

The text, the message does not imply object creation, database manipulation or anything. It is just… counting stuff.

cacarrara: So, the output says
> Current count: 42 reimbursements

But `Reimbursement.objects.count()` says 38, for example. This is a misleading information for who import Reimbursements.

An output like this would be clearer:
> Processed: 42 reimbursements
> Stored: 38 reimbursements
> Failed: 4 reimbursements

cuducos: Wow… now I got it. Great idea! Probably there are tons of stuff to count: lines processed, actual count in the DB, how many records were included, how many were updated.

> edit: probably it's a new feature. We shouldn't hold this PR because of this.

Would you mind opening an issue with a summary so we work on that later?

jtemporal: >probably it's a new feature

agree on that! =)

jtemporal:
I believe we are ready to merge =)

So I'm merging this one.

Some others relevant points were raised here like:
- using model_mommy
- improving the feedback to user

I believe opening up issues is a way to proceed.

Thank you all!

cuducos:
> using model_mommy

No need to use `model_mommy` after #171 — tks @cacarrara ; )

> improving the feedback to user

Yep, this is a good idea ; )

PR #171, opened 2017-05-26
Use Mixer for reimbursement view tests

cacarrara:
This change makes use of mixer[1] for fixtures in Reimbursement view tests.

[1] - https://github.com/klen/mixer/


cacarrara:
@cuducos take a look. More than a pull request this is a proposal to improve tests.

.travis.yml
```diff
@@ -37,6 +37,7 @@ install:
   - psql -U postgres -c 'create database "jarbas";'
   - yarn install
   - python -m pip install -r requirements.txt coveralls
+  - python -m pip install -r requirements-test.txt coveralls
```
cuducos: We're trying to install `coveralls` twice. I'd change the strategy here: 

* Rename `requirements-test.txt` to `requirements-dev.txt`
* Include `-r requirements.txt` as the first line of in the new `requirements.txt`
* Move `django-test-without-migrations` from `requirements.txt` to `requirements-dev.txt`
* Use _only_ `requirements-dev.txt` in `.travis.yml`

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -1,59 +1,19 @@
-from datetime import date
 from json import loads
 from unittest.mock import patch
+from urllib.parse import urlencode
 
-from django.core.cache import cache
 from django.shortcuts import resolve_url
 from django.test import TestCase
```
cuducos: Unnecessary line: mixer can be imported in the sabe block of Django import (third party imports, according to PEP8)

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -1,59 +1,19 @@
...
-        data[4]['subquota_id'] = 22
-        data[4]['year'] = 1983
-        data[4]['issue_date'] = '1960-02-01'
-        del data[4]['probability']
-        del data[4]['suspicions']
-
-        for fixture in data:
-            Reimbursement.objects.create(**fixture)
-
+        mixer.cycle(5).blend(Reimbursement)
```
cuducos: Actually the 5 reimbursements created before had controlled differences to make the tests more precise. Surely the previous code was poor and yours is way better. So… just to confirm: `5` here is quite arbitrary (before we need five due to the combination of differences to properly test API requests), right? I mean… when we need actually with a specificity now we create on the test itself, not during `setUp`, is that right? If so we might reduce from 5 to… whatever, 3. What do you think?

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -67,38 +27,43 @@ def test_content_general(self):
...
+        self.assertTrue(first['issue_date'] > last['issue_date'])
+
+    def test_content_with_cnpj_cpf_filter(self):
+        search_data = (
+            ('cnpj_cpf', '12345678901'),
+            ('subquota_id', '22'),
+            ('order_by', 'probability'),
+            ('suspicious', '1'),
         )
+        url = '{}?{}'.format(self.url, urlencode(search_data))
```
cuducos: ❤️

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -1,59 +1,19 @@
...
+from urllib.parse import urlencode
 
-from django.core.cache import cache
 from django.shortcuts import resolve_url
 from django.test import TestCase
 
+from mixer.backend.django import mixer
+
 from jarbas.core.models import Reimbursement
-from jarbas.core.tests import sample_reimbursement_data, suspicions
```
cuducos: If we're replacing this fixture, we can cleanup `jarbas/core/tests/__init__.py` and use `mixer` in other tests that depends on them… using `mixer` _sometimes_ IMHO tends to lead to a bigger mess ; )

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -123,76 +89,52 @@ def test_status(self):
...
-            probability=0.5,
-            suspicions=suspicions,
-            receipt=dict(fetched=False, url=None)
-        )
-        self.assertEqual(expected, contents)
+        for result_attr, result_value in contents.items():
+            if not hasattr(self.reimbursement, result_attr):
+                continue
+            expected_value = getattr(self.reimbursement, result_attr)
+            self.assertEqual(str(result_value), str(expected_value))
```
cuducos: This logic is pretty clever but I think it makes it less precise what we're are actually testing… IMHO it's a good idea to keep logic extremely simple in tests to make them obvious. That's why here I prefer the long `dict`: in a glimpse one can know what to expect as result of `test_contents`, the proper types etc. What do you think?

jarbas/api/tests/test_reimbursement_view.py
```diff
@@ -123,76 +89,52 @@ def test_status(self):
...
         resp = self.client.get(self.url)
-        expected = dict(url=self.expected_receipt_url)
+        expected = dict(url=self.reimbursement.receipt_url)
         content = loads(resp.content.decode('utf-8'))
         self.assertEqual(expected, content)
 
     @patch('jarbas.core.models.head')
     def test_fetch_non_existing_receipt(self, mocked_head):
         mocked_head.return_value.status_code = 404
-        cache.clear()
```
cuducos: Any specific reason to remove that?

jarbas/core/models.py
```diff
@@ -116,7 +116,11 @@ def as_list(content, cast=None):
             return None
 
         parts = list(content.split(','))
-        return list(map(lambda x: cast(x), parts)) if cast else parts
+        try:
+            return list(map(lambda x: cast(x), parts)) if cast else parts
+        except ValueError:
+            # this is far from good
+            return list()
```
cuducos: I'm assuming the `except ValueError` might happen in `cast`, right? If so, I don't think we should silence this error. If the user is trying to convert something unconvertible it's better to stop temh than to loose data silently.

jarbas/api/tests/test_reimbursement_view.py
cacarrara: @cuducos, good points. I'll gona try do not be too extense.

I do not agree completly about that. The test here should test that each API response attribute is equal to the object attribute in database. It's what being done properly. Of course:

```
response = load(api_response)
mydict = dict(one=1, two=2)
assert mydict == mydict
```

is in a first glance such more obvious than:

```
response = load(api_response)
mydict = dict(one=1 two=2)
for key, value in mydict.items():
	return assert response[key] == value
```

However I still think the positive consequences are higher than negative ones. Think about changing (for any reason) the response content structure (not values). If we change an attribute name or even any attribute data type. We'll have to adapt the tests only because we change some implementation specificity, not a business rule, nor an application requirement. On the other hand, if we change some business logic that makes API response different from object attribute value, the test still will fail.

I think the tests should be abstract and resilient enought to test requirements properly without being a pain for developers.

It's just my thoughts, let me know what you think.

ps.: about this topic, I would recomend this post: http://blog.cleancoder.com/uncle-bob/2017/05/05/TestDefinitions.html

jarbas/api/tests/test_reimbursement_view.py
cacarrara: I had understood it was only preparing test to fetch an exiting receipt from a reimbursement without a receipt. So I did the proper preparation instead of clearing cache, what seemed to be an workaround. Since now is easy and simple create new objects I did it. 

Was there any other reason to clear the cache?

cacarrara:
@cuducos do you think we should change all tests within only one PR? I think we could do it one PR by test file. It could be easier and better to review/approve so we could minimize chances for errors.

Have temprorarily only some tests running with mixer seems do not raise any issue.

I can handle the tests refactoring.

jarbas/api/tests/test_reimbursement_view.py
cuducos: >  If we change an attribute name or even any attribute data type. We'll have to adapt the tests only because we change some implementation specificity, not a business rule, nor an application requirement.

If we change an attribute name we broke the JSON API. So… yep, I think a failing test would be welcomed in that scenario ; )

jarbas/api/tests/test_reimbursement_view.py
cuducos: Hum… if for some reason the URL was accessed before the cache might fools the test. As we have no control of tests ordering, this might happen… but it's not a sure thing. Let's remove it and if this become a problem we put it back ; )

jarbas/api/tests/test_reimbursement_view.py
cuducos: In addition to my former comment: 

The test you're suggesting is indeed interesting, but as a `test_context` (i.e. testing the data we're getting from the model and passing to the template). But with the class-based view from DRF this step is abstracted (and consequentially not tested).

What `test_contents` is actually testing is if the JSON output is what our frontend expects. So even if we change some attribute name, we can do some tweaks to avoid breaking compatibility with the frontend (i.e. tests will pass because JSON won't change) — or, surely better, have a failing test to remind us we've broken something and the JSON is not what we expected it to be anymore, so we can adapt the frontend and our expectations….

cacarrara: I got your point and you're right. I was considering the `test_content` as an unit test, but it's an integration test actually. I'm going to do it.

cacarrara:
@cuducos and @jtemporal the changes was done. Take a look :stuck_out_tongue_winking_eye:

cuducos:
Wow! Many thanks, @cacarrara <3

PR #173, opened 2017-05-27
Enhance user output when loading reimbursements

cuducos:
As suggested by @cacarrara in #170:

### DIFF
--- a/jarbas/core/management/commands/reimbursements.py
+++ b/jarbas/core/management/commands/reimbursements.py
@@ -2,6 +2,7 @@ import csv
 import lzma
 
 from django.utils.timezone import now
+from reprint import output
 
 from jarbas.core.management.commands import LoadCommand
 from jarbas.core.models import Reimbursement
@@ -13,15 +14,18 @@ class Command(LoadCommand):
     def handle(self, *args, **options):
         self.started_at = now()
         self.path = options['dataset']
-        self.count = Reimbursement.objects.count()
-        print('Starting with {:,} reimbursements'.format(self.count))
+
+        exixting = Reimbursement.objects.count()
+        print('Starting with {:,} reimbursements'.format(exixting))
+        self.count = {'updated': 0, 'created': 0, 'skip': 0}
 
         if options.get('drop', False):
             self.drop_all(Reimbursement)
-            self.count = 0
 
-        self.create_or_update(self.reimbursements)
-        self.print_count(Reimbursement, count=self.count, permanent=True)
+        with output() as status:
+            status.change(self.status)
+            self.create_or_update(self.reimbursements, status)
+
         self.mark_not_updated_reimbursements()
 
     @property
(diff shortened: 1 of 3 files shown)


jarbas/core/management/commands/reimbursements.py
```diff
@@ -13,15 +14,18 @@ class Command(LoadCommand):
     def handle(self, *args, **options):
         self.started_at = now()
         self.path = options['dataset']
-        self.count = Reimbursement.objects.count()
-        print('Starting with {:,} reimbursements'.format(self.count))
+
+        exixting = Reimbursement.objects.count()
```
cacarrara: There's a typo (exixting) or this word is new for me. :smile:

cuducos: Hahaha… typos is one of my skills! Fixed in b4978d9

jarbas/core/management/commands/reimbursements.py
```diff
@@ -15,7 +15,7 @@ def handle(self, *args, **options):
         self.started_at = now()
         self.path = options['dataset']
 
-        exixting = Reimbursement.objects.count()
+        existing = Reimbursement.objects.count()
         print('Starting with {:,} reimbursements'.format(exixting))
```
jtemporal: dear @cuducos you forgot to correct the typos :P `exixting` lol

cuducos: Ooops… done.

jarbas/core/management/commands/reimbursements.py
```diff
@@ -77,16 +81,36 @@ def serialize(self, reimbursement):
...
     def mark_not_updated_reimbursements(self):
         qs = Reimbursement.objects.filter(last_update__lt=self.started_at)
         qs.update(available_in_latest_dataset=False)
+
+    @property
+    def status(self):
+        label = '{}s'.format(self.get_model_name(Reimbursement)).lower()
+        total = sum(map(self.count.get, self.count.keys()))
+        output = [
+            'Processed: {} lines'.format(total),
```
jtemporal: Why `lines` instead of `reimbursements`? I know each line is one reimbursement but I think would be clearer if we said reimbursements instead of lines.

cuducos: Hum… I coded it that way because I wasn't considering the CSV line a reimbursement until the [_has a `document_id`_ check](https://github.com/datasciencebr/jarbas/blob/3b9a0b923159187ce52f0068be7d4ddc780889a8/jarbas/core/management/commands/reimbursements.py#L88-L91). Not sure if it makes sense — how do you prefer?

jtemporal: It totally makes sense! I knew I was forgetting to take into account something =P

cacarrara: What if implement status() property on LoadCommand class and just define the output here as an attribute? Seems it would be good have this method available for other commands.

cuducos: Sounds good… feel free to register the idea in a new _Issue_!

Issue #175, opened 2017-05-28
Create API endpoint for reimbursements that weren't available on the last dataset


jtemporal:
Since #170 created a flag `available_in_latest_dataset` maybe we should have an endpoint that has all reimbursements where that flag is set to `False`?

jtemporal:
closed by #176

PR #176, opened 2017-05-28
Add “deleted” reimbursement filter to the API

cuducos:
As suggested in #175


README.md
```diff
@@ -61,9 +61,12 @@ All these endpoints accepts any combination of the following parameters:
 * `issue_date_end` (exclusive)
 * `month`
 * `subquota_id`
-* `suspicions` (_boolean_, basically any value parses to `True`)
+* `suspicions` (_boolean_, `1` parses to `True`, `0` to `False`)
 * `year`
-* `order_by`: `issue_date` (default) or `probability` (both descending)
+* `order_by`: `issue_date` (default) or `probability`
+(both descending)
```
jtemporal: Is this line break intentional or a typo?

jarbas/api/views.py
```diff
@@ -26,9 +26,16 @@ def get(self, request):
...
 
-        # filter queryset
-        if self.request.query_params.get('suspicions'):
+        # filter suspicions
+        if self._bool_param('suspicions'):
             self.queryset = self.queryset.suspicions()
+
+        # filter reimbursement in latest dataset
+        in_latest = self._bool_param('in_latest_dataset')
+        if in_latest is not None:
```
jtemporal: Quick question: Why do we use `None` instead of a condition containing `True` or `False`? Maybe I got lost on the logic here =)

jarbas/core/tests/test_reimbursement_model.py
```diff
@@ -89,6 +89,17 @@ def test_suspicions(self):
         Reimbursement.objects.create(**self.data)
         self.assertEqual(1, Reimbursement.objects.suspicions().count())
 
+    def test_in_latest_dataset(self):
```
jtemporal: I would split this test in two: One for the deleted and one for the not-deleted. Even though they are part of the same logical thinking of testing whether or not the reimbursement is available in the latest dataset, it seems to me that breaking this test in two would make easier to detect code failure in a near future.

jarbas/api/views.py
cuducos: * `None`: user haven't user this filter
* `False`: user have used this filtering explicitly requesting records where `available_in_latest_dataset` is `False` (i.e. reimbursements deleted in the latest version of the Chamber of Deputies dataset)
* `False`: user have used this filtering explicitly requesting records where `available_in_latest_dataset` is `True` (i.e. reimbursements that are still in the latest version of the Chamber of Deputies dataset)

README.md
cuducos: Not necessarily a typo because in `.md` this line break has no effect — but surely it wasn't intentional. Removed in 9723951.

jarbas/core/tests/test_reimbursement_model.py
cuducos: Good idea: done in cc8f421.

jarbas/api/views.py
jtemporal: hmmmm Thanks!

cuducos:
Ooops… saw the failing test. Gonna check it… just a minute.

cuducos:
Fixed in e9f5f99

jtemporal:
>Ooops… saw the failing test

Oooh the beauty of automatically testing ❤️

Issue #174, opened 2017-05-27
Track reimbursement changes


cuducos:
Putting together #170 and [a tweet](https://twitter.com/leleportella/status/868561613191868416) from @leportella that introduced me to [`django-simple-history`](https://django-simple-history.readthedocs.io/) I think it's a good idea to use it to track reimbursement changes. (Thanks Leticia for the unpredictable help!)

leportella:
Nice!

PR #177, opened 2017-05-28
Add django-simple-history to keep track of Reimbursement changes

cuducos:
As suggested in #174 (and added Django Admin — for admins only, by now — to better visualize the new feature)

image


jarbas/core/admin.py
```diff
@@ -0,0 +1,57 @@
...
+        'state',
+    )
+
+    def is_suspicious(self, obj):
+        return obj.suspicions is not None
+
+    is_suspicious.short_description = 'Suspicious'
+    is_suspicious.boolean = True
+
+
```
jtemporal: I thought you were going to show the historical records in the admin. And if you weren't why not?

cuducos: We are:

jtemporal: Oooh I see it now, damn top right button!

cuducos:
> Also I would add to README the ref to Django Docs were it shows the `createsuperuser` command

Good idea! Done in 42d29c7

PR #180, opened 2017-05-29
Public dashboard

cuducos:
This PR adds a public dashboard (no login needed) to Jarbas so people have an alternative way to browse a list of reimbursements:

It will be available at `jarbas.serenatadeamor.org/dashboard/` and no creation, update or delete actions are permitted.


jarbas/dashboard/sites.py
```diff
@@ -0,0 +1,84 @@
...
+            permission = Permission.objects.get(
+                content_type=ContentType.objects.get_for_model(Reimbursement),
+                codename='change_reimbursement'
+            )
+            user.user_permissions.add(permission)
+            user = user_model.objects.get(username='dashboard')  # avoid cache
+        return user
+
+    @staticmethod
+    def valid_url(url):
```
henriquebastos: Why is this necessary? Restricting to GET isn't enougth?

jarbas/dashboard/sites.py
```diff
@@ -0,0 +1,84 @@
...
+    site_title = 'Dashboard'
+    site_header = 'Jarbas Dashboard'
+    index_title = 'Jarbas'
+
+    def __init__(self):
+        super().__init__('dashboard')
+        self._actions, self._global_actions = {}, {}
+
+    @staticmethod
+    def get_user():
```
henriquebastos: Criar esse usuário _dashboard_ é realmente necessário? Implementar o `has_permissions` filtrando `GET` e liberando o resto não é suficiente?

jarbas/dashboard/templates/admin/base.html
```diff
@@ -0,0 +1,71 @@
+{% load i18n static %}<!DOCTYPE html>
```
henriquebastos: Pq vc precisou replicar todos os htmls? Não dá pra saber a diferença do original olhando o commit.

cuducos:
Many many thanks @henriquebastos!

After a hacking session with @riverfount we could get rid of the real `User` model, envolving database etc. Now an `AnonymousUser` handles the task pretty well.

TBH I also felt uncomfortable customizing three HTML files. What I would like to do is to rmeove certain elements of the UI that have no meaning in this hacked readonly loginless version:

---

pasted_image_31_05_17_19_09

---

pasted_image_31_05_17_19_14

---

Any ideas of a more effortless (thinking of medium/long term maintenance) to handle this?

henriquebastos:
Really liked the end result.

I would tweak a couple things:

1. Simplify tests or make them more concise avoiding setup inside a test
method.
2. Replace map and filter with generator expressions.
3. Improve naming on purpose instead of content. (like the PostPut test
case that actually verifies forbidden methods).

Not that it seems necessary, just pointing out for the fun of it. :-P

cuducos:
Many thanks again, @henriquebastos! Gonna keep that in mind and open some PR with these minor edits soon ; )

PR #183, opened 2017-06-02
Add requeriments-dev.txt to docs & Dockerfile

cuducos:
As documentation and Dockerfile target developers, this sounded reasonable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM python:3.5
 COPY requirements.txt /requirements.txt
+COPY requirements-dev.txt /requirements-dev.txt
 RUN python -m pip install -U pip
-RUN python -m pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM python:3.5
 COPY requirements.txt /requirements.txt
```
jtemporal: why do you copy the `requirements.txt` but doesn't run it?

cuducos: Because [`requirements-dev.txt` depends on `requirements.txt`](https://github.com/datasciencebr/jarbas/blob/master/requirements-dev.txt#L1).

jtemporal: Maaaann forgive my lack of attention! I are correct! Sorry ;)

cuducos: C'mon… no worries — that's what code reviews are for ; )

PR #185, opened 2017-06-02
Restart server on post-receive

cuducos:
This simple script is used as a base for our deploy. Automating the process of restarting `gunicorn` is a simple enhancement. 

**Disclaimer**: There is minor cleanup hidden in this PR. I run into a script that was only used in an older version of `Dockerfile` so I deleted it.


jtemporal:
LGTM

Irio:
@cuducos Can you solve the conflicts so it can be merged?

cuducos:
Done!

PR #187, opened 2017-06-02
Add Google Analytics to dashboard

cuducos:
Following DRY now Google Analytics is in a context processor, so both the Elm frontend and the dashboard get the value.

### DIFF
--- a/jarbas/core/context_processors.py
+++ b/jarbas/core/context_processors.py
@@ -1,0 +1,5 @@
+from django.conf import settings
+
+
+def google_analytics(request):
+    return {'google_analytics': settings.GOOGLE_ANALYTICS}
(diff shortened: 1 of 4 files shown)


jtemporal:
As far as the dashboard goes, the GA script is working perfectly but:

1. I'm not the greatest on analytics but I thought it should be placed as close to `head` tag as possible
1. It is not showing on the `Elm` frontend yet

on 2. is that going to be fixed on a later commit or new PR?

cuducos:
> I'm not the greatest on analytics but I thought it should be placed as close to head tag as possible

Old habits here: I place it in the bottom because it was supposed to be the last thing the browser should care about.

> It is not showing on the Elm frontend yet

Can you confirm? Here, running with a _envvar_ (eg ` GOOGLE_ANALYTICS=xuxa python manage.py runserver` and yes I used Xuxa for testing).

jtemporal:
> Can you confirm? 

It works, it was me again testing out things wrongly again 🙄 

Merging it =)

PR #192, opened 2017-06-07
Allow users to order by value column

cuducos:
Basically a one line commit to allow users to sort results in the dashboard using the _value_ column:

pasted_image_07_06_17_15_19

Thanks @luzfcb for the solution. Thanks @vncscoelho for the [bug report](https://twitter.com/vcoelho/status/872495947841982468).

### DIFF
--- a/jarbas/dashboard/admin.py
+++ b/jarbas/dashboard/admin.py
@@ -248,6 +248,7 @@ class ReimbursementModelAdmin(SimpleHistoryAdmin):
         return 'R$ {:.2f}'.format(obj.total_net_value).replace('.', ',')
 
     value.short_description = 'valor'
+    value.admin_order_field = 'total_net_value'
 
     def still_available(self, obj):
         return obj.available_in_latest_dataset


luzfcb:
in a future, maybe replace all money fields (`DecimalField`) to `MoneyField` ( http://django-money.readthedocs.io/en/stable/)

This would facilitate calculations if there are expenditures in different currencies

cuducos:
> This would facilitate calculations if there are expenditures in different currencies.

Jarbas doesn't calculate anything so far. It's basically an API to show what the Chamber of Deputies shows — but with enhanced features to facilitate auditioning. So we basically calculate nothing, just show stuff to users. By the way we even replicate some arguably inconsistente data such as `year` with a different year than `issue_date`, for example.

Also our database is only in Brazilian Reais — even expenditures made abroad are converted by congresspeople before claiming for the reimbursement. We have rolled out a feature to help users check if the congressperson's converted currency properly in #75.

Therefore I see no use ATM for `django-money`. It seams so powerful to only help in… sorting fields. But is surely is something to keep in mind ; )

anaschwendler:
LGTM ;)

PR #193, opened 2017-06-09
Fix url from navigation bar

cassiobotaro:
http://serenata.serenatadeamor.org/ -> https://serenatadeamor.org


cuducos:
Ooops… tks ; )

PR #197, opened 2017-06-19
Use django_extensions in debug mode

cuducos:
Sometimes I like to check how Django ORM talks to the database (i.e. see the SQL queries). This extension allows something like:

```
$python manage.py shell_plus --print-sql
```

And then:

```
In [3]: Reimbursement.objects.filter(year=2017).exclude(suspicions=None).count()
SELECT COUNT(*) AS "__count"
FROM "core_reimbursement"
WHERE ("core_reimbursement"."year" = 2017
       AND NOT ("core_reimbursement"."suspicions" IS NULL))


Execution time: 0.070313s [Database: default]

Out[3]: 46
```


cuducos:
Dear code reviewer: Code Climate red badge is fixed by #208 (in which we configure larger line length than PEP8).

PR #203, opened 2017-06-21
Clean up node_modules and elm-stuff & update pip in each deploy

### LINKED ISSUE #200: Delete elm-stuff and node_modules in the beginning of each deploy

cuducos:
Fix #200


contrib/post-receive
```diff
@@ -12,9 +12,16 @@ echo "\n==> Activating virtualenv…\n"
...
+    rm -rf node_modules/
+fi
+if [ -d elm-stuff/ ]; then
+    rm -rf elm-stuff/
+fi
 yarn install
 
 echo "\n==> Installing Python packages…\n"
+pip install -U pip
 pip install -r requirements.txt
```
jtemporal: wouldn't it be case to add a `-U` on here too? by forcing the update of outdated packages?

cuducos: That's an idea, but… `requirements.txt` _should_ have precise versions, not `wahtever>=1.0.0`.

That said, I don't think it's _necessary_, but I won't oppose this change.

What about you: _that said_, can I add the `-u` or maybe I have just convinced you otherwise?

jtemporal: I like the `-U` because sometimes simply running `pip install -r requirements.txt` did not update some libs I had in a previous version even when requirements had an higher version.

cuducos: Indeed! Done in 32d4a65

jtemporal:
Today I'll finalize Jarbas' provisioning documentation. As soon as that is done, I'll get back to this here on the same machine :)

Issue #205, opened 2017-06-21
Enhance security 


cuducos:
@davinirjr alerted my that we are [ranked D+ in terms of security](https://observatory.mozilla.org/analyze.html?host=jarbas.serenatadeamor.org). Is there a way to improve that?

davinirjr:
What whe use on the front? Nginx or Apache?

cuducos:
Nginx — the base config is [`contrib/nginx`](https://github.com/datasciencebr/jarbas/blob/master/contrib/nginx).

davinirjr:
Awesome! I'll help on it.

PR #207, opened 2017-06-22
Upgrade SSL grade on Mozilla Observatory

davinirjr:
https://observatory.mozilla.org/analyze.html?host=jarbas.serenatadeamor.org

### DIFF
--- a/contrib/nginx
+++ b/contrib/nginx
@@ -1,10 +1,28 @@
 server {
-    listen              443 ssl;
+    listen              443 ssl http2;
+    listen              [::]:443 ssl http2;
     server_name         jarbas.serenatadeamor.org;
     ssl_certificate     /etc/letsencrypt/live/jarbas.serenatadeamor.org/fullchain.pem;
     ssl_certificate_key /etc/letsencrypt/live/jarbas.serenatadeamor.org/privkey.pem;
-    ssl_protocols       TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers         HIGH:!aNULL:!MD5;
+    ssl_protocols       TLSv1.2;
+    ssl_ecdh_curve secp384r1;
+    ssl_prefer_server_ciphers on;
+    ssl_session_timeout 1d;
+    ssl_session_cache shared:SSL:10m;
+    ssl_session_tickets off;
+    #ssl_ciphers         HIGH:!aNULL:!MD5;
+    #ssl_ciphers "EECDH+AESGCM:EECDH+AES";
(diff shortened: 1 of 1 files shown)


jtemporal:
Hi @davinirjr thank you for your PR! But I'm having a little trouble evaluating it, would you mind providing a better description of what was your idea here? :)

davinirjr:
No problem at all, @jtemporal.

Today, we have D+ on Mozilla's Observatory:

image

@cuducos and I talked about it, and I just helped on providing a more secure SSL config for Jarbas site.

He pointed that we could verify the changes on test environment that you're working on (since we need real certificates, and for Let's Encrypt, we need a fully functional public-DNS host).

davinirjr:
Better?

cuducos:
> Hi @davinirjr thank you for your PR! But I'm having a little trouble evaluating it

Hi @jtemporal, I'm sorry I missed that: I had a ~1h call with @davinirjr in which he explained line by line this nginx confs. Merging it here doesn't change anything in production (it only properly document the settings we're using).

So what's next? You or I should log into the production server, add the confs line by line testing its effect. If we can push all of them to production, then I consider this ready to merge. Does that make sense?

jtemporal:
> Better?

Yes!

> he explained line by line this nginx confs

For future reference: That would be nice to have reflected on the commit messages

>  Does that make sense?

It does.

I still have concerns on testing this in production though. 

Another question (I'm really green on nginx and all of that): is there a reason why there many commented cipher lines in there?

davinirjr:
> I still have concerns on testing this in production though.

Huge +1 here. 

But to verify this PR we'll need a "live" server, aka, public DNS with certificates, as Observatory will not reach internal IP's and Let's Encrypt will not generate certificates either.

Do we have an live QA server with certificate? We can measure it before and after to make sure this will not break anything.

> Another question (I'm really green on nginx and all of that): is there a reason why there are many commented cipher lines in there?

If we use better the ciphers, we choose to drop out "older" browser support.

I don't think we could push and test line by line, as @cuducos says (there are lines that have some dependencies), but I can help on it too. 

And nginx can test the conf file before it loads into running instance, using -t in cli:

https://www.nginx.com/resources/wiki/start/topics/tutorials/commandline/

jtemporal:
Just to link it: This is supposed to fix #205

jtemporal:
@filipecifali showed me [this other SSL scanner](https://www.ssllabs.com/ssltest/analyze.html?d=jarbas.serenatadeamor.org) the other day that showed a not so low score for Jarbas. Thought you guys might find it interesting =)

davinirjr:
Yep, Mozilla Observatory uses SSLLabs on "third-party" tests tab.

PR #208, opened 2017-06-23
Add more engines to Code Climate

cuducos:
Testing new possibilities for Code Climate.

WIP because I'm gonna test it as a coverage CI as well.

PR #212, opened 2017-06-28
Add link to Rosie's tweets

### LINKED ISSUE #209: Add link to @RosieDaSerenata tweets
A possible roadmap would be a custom Django command to scrap @RosieDaSerenata to:

- [x] Parse tweets and figure out the `document_id`<br>done in 65af517 @  [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Check if tweet exists in Jarbas database<br>done in 6736039 @  [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Save link to a new model (or add a field to `Reimbursement` model)<br>done in 6736039 @  [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Write tests<br>done in f462cb1 & 6820cf0 @  [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Add info to JSON API view<br>done in df042cb @  [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Add a link in the UI (eg next to fetch/view receipt )<br>done in b7c145a @ [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)
- [x] Add links in the dashboard<br>done in 4453e6d @ [`cuducos-read-rosies-tweets`](https://github.com/datasciencebr/jarbas/tree/cuducos-read-rosies-tweets)

cuducos:
Fix #209:

* Implements Django custom command to read Rosie's tweets and identify the linked Reimbursement
* Adds link to Rosies tweets to the dashboard and to the Elm front-end, as well as tweet URL to the JSON API

screen shot 2017-06-28 at 10 50 59

screen shot 2017-06-28 at 10 52 47


cuducos:
Dear code reviewer: Code Climate red badge is fixed by #208 (in which we configure larger line length than PEP8).

README.md
```diff
@@ -281,4 +279,14 @@ If **you are not** using Docker copy `contrib/.env.sample` as `.env` in the proj
 ##### Google settings
 
 * `GOOGLE_ANALYTICS` (_str_) Google Analytics tracking code (e.g. `UA-123456-7`)
-* `GOOGLE_STREET_VIEW_API_KEY` (_str_) Google Street View Image API key
\ No newline at end of file
+* `GOOGLE_STREET_VIEW_API_KEY` (_str_) Google Street View Image API key
+
+##### Twitter settings
```
jtemporal: Would be nice to document the `tweets` command just like we have documented the ones that load data into the database.

cuducos: Indeed! Added by 132d82c

jtemporal:
FYI: I can't get my machine to show both pretty changes so I'll pair with @cuducos later to check if there's something missing

jtemporal:
We have a 🐛 \o/

Interface wise everything is working ( @cuducos and I tested by add a tweet manually on the `shell`) but creating the links by running `python manage.py tweets` does not persist the changes in the database 😢 as we can see:

screen shot 2017-07-06 at 1 32 53 pm

1. First we need `BigIntegerField` to store tweets `IDs`
2. Then an ideia is to use: `Tweet.objects.create(reimbursement=reimbursement, status=status)` instead of `.save()`

cuducos:
Many thanks for the careful review, @jtemporal : )

> running `python manage.py tweets` does not persist the changes in the database

Fixed in a9346cb

> First we need `BigIntegerField` to store tweets IDs

Fixed in 3fa680f

It should be working now:

```
In [1]: from jarbas.core.models import Tweet

In [2]: Tweet.objects.count()
Out[2]: 111
```

cuducos:
Reverted 3fa680f because tweet status are already in the `10**18` scale and `BigInteger` allows up to `10**19`. So I used `Decimal` with `0` decimals and max length of 25 (i.e. we can store numbers up to `10**25` _ and it's quite easy to change this number editing `models.py` and generating a migration automatically).

jtemporal:
hey @cuducos awesome bug squashing!!!

Buuuut... tests are failing:
screen shot 2017-07-07 at 6 08 17 pm
would ou mind checking it out? :D

cuducos:
That's a bizarre bug! I fixed, but let me go through it:

1. We use a mixer [mixer](https://github.com/klen/mixer/) to generate some fixtures in testsand mixer uses [faker](https://github.com/joke2k/faker)
1. We're using `DecimalField` not because we need decimal numbers, but because we need to store numbers higher than `10**19` (Django ORM's `BigInteger` is not enough)
1. In other words these big numbers don't have a decimal part, I just picked `DecimalField` because they wouldn't fit `BigInteger` (they are set with [`decimal_places=0`](https://github.com/datasciencebr/jarbas/blob/dbfe66244e51f50f5b235ec3d7956908d8d11872/jarbas/core/models.py#L182))
1. Python don't allow `Decimal` with zero precision (try `$ python -c "import decimal; decimal.getcontext().prec = 0"` and you'll get `ValueError: valid range for prec is [1, MAX_PREC]`)
1. Therefore faker [won't generate a decimal with zero decimal places](https://github.com/joke2k/faker/blob/3f78b0f6975616dae47b0b625921aa2566c2bec9/faker/providers/python/__init__.py#L46-L48)
1. And finally any attempt to `mixer.blend(Tweet)` will fail because no value will be generated for `Tweet.status` field

My solution was to always passa a value to `Tweet.status`, even we don't need one — that's 3a135d6.

jtemporal:
hey @cuducos as said before to you I had some concerns on creating a random tweet status... So I went looking for an alternative. And found [this awesome blog post from @mattwaite that might be of help]( http://www.mattwaite.com/posts/2009/mar/10/django-and-really-big-numbers/) want to talk about the implications of this over the next days ;)

cuducos:
@jtemporal I'm not sure I've got your point. This blog post is from 2009. Nowadays, 8 years later, [Django ORM actually have `BigInteger`](https://docs.djangoproject.com/en/1.11/ref/models/fields/#bigintegerfield) and indeed I attempted to use it in this PR (3fa680f). The problem is that `BigInteger` might not be big enough as I explained below — the message you reacted with a 👍.

So… what's the idea? Adapt a [`DecimalField`](https://docs.djangoproject.com/en/1.11/_modules/django/db/models/fields/#DecimalField) for our use case? That could work… but… the hack to generate a random tweet in tests would still exists: this is a Faker problem, not Django or Mixer…

antonioj-mattos:
@cuducos 

have you consider using the `id_str` instead, given this concern regarding the range of int64?

see the recomendation [on this post](https://blog.twitter.com/developer/en_us/a/2013/64-bit-twitter-user-idpocalypse.html) about regarding using `id`s

> Where available, “id_str” continues to be the best way to access any ID in the system, even for platforms which support 64-bit integers.

cuducos:
> have you consider using the `id_str` instead

@antonioj-mattos in fact yes, I have. Just considered. But as our database is already struggling with performance I decided to stick to integers and save a bit of processing here. Not sure if it makes a lot of sense —lookups generally comes from the `Reimbursement` instance, not the `Tweet.status` _per se_… 

…  as I'm a noob in database issues I thought  keeping numbers in numeric fields would mean something. Feel free to correct me if I'm wrong ; )

antonioj-mattos:
@cuducos 

> I thought keeping numbers in numeric fields would mean something

sure, it saves up to 5x memory space for the indexes, which by itself might be a good reason ;)

> our database is already struggling with performance

I probably would not have asked if I'd known this before. Although, for this use case the extra processing may be negligible, you're right as performance problems pile up.

cuducos:
Many thanks for the input anyway, @antonioj-mattos! Something that is still pending is a review of our models to be more kind with Postgres — but this is not registered anywhere, our fault. I'm sorry about it.

antonioj-mattos:
@cuducos 

Sorry to bother you again. But I was reading Twitter's API docs and have some thoughts, one of which:

 - Would it be possible to invert the processing ordering? Since the response gives tweets from the most recent to the least recent. That way we would always have the most recent status on the last row, as such we could eliminate the index on the status field and use de `id` directly ie. `where id = max(id)`. Then having no need for search by the status we could make it a string.

I am just not sure about the response order guarantees: if we call the endpoint twice will it return the tweets in the same order?

PR #214, opened 2017-06-30
Apply minor simplifications

antonioj-mattos:


jarbas/core/models.py
```diff
@@ -124,8 +124,8 @@ def as_list(content, cast=None):
         if not content:
             return None
 
-        parts = list(content.split(','))
-        return list(map(lambda x: cast(x), parts)) if cast else parts
+        parts = content.split(',')
+        return list(map(cast, parts)) if cast else parts
```
cuducos: What about `return [cast(p) for p in parts] if cast else parts`?

jarbas/core/management/commands/irregularities.py
```diff
@@ -62,7 +62,7 @@ def serialize(self, row):
         hypothesis = tuple(k for k in row.keys() if k not in reserved_keys)
         pairs = ((k, v) for k, v in row.items() if k in hypothesis)
         filtered = filter(lambda x: self.bool(x[1]), pairs)
-        suspicions = dict((k, True) for k, v in filtered) or None
+        suspicions = {k: True for k, v in filtered} or None
```
cuducos: What about `{k: True for k, _  in filtered}` as we're not using `v`?

jarbas/core/models.py
antonioj-mattos: @cuducos it is up to you, my strategy was that of changing as little as possible

cuducos: Once we're enhancing it, let's enhance it as far as we can ; )

PR #225, opened 2017-07-28
Load reimbursements text

### LINKED ISSUE #222: Import receipts texts (from OCR)
Adding OCR data (`2017-02-15-receipts-texts.xz`) to `Reimbursement` and adding it to the dashboard would make @tatianasb and @vilapedro lives easier: they could search in receipt text easily (without depending on any tech interface).

A possible roadmap:

1. Add a field to `Reimbursement` model
1. Add a command similar to the one that load suspicions
1. Add this field (and set is as searchable) in the admin/dashboard

jtemporal:
fix #222


jtemporal:
Missing steps:

- [x] Add a field to Reimbursement model
- [x] Add a command similar to the one that load suspicions
- [ ] Add this field (and set is as searchable) in the dashboard

Also, this new command has **a lot** of testing similar to suspicions, some testing refactor is required ;)

Issue #178, opened 2017-05-28
Use Mixer for tests


cuducos:
As @cacarrara started in #171 ; )

The idea is to use this tool in all other test that needs fixtures.

cacarrara:
I suggest we could do it one test file by time. It would be easier to review and less dangerous :smile: -- I could do it.

cacarrara:
You know how to use this Milestones feature? We could use it.

cuducos:
> You know how to use this Milestones feature?

The feature is pretty cool, but in terms of community management I had a not-so-good experience with it. It is not as visible as _Issues_, so it ends up as a kind of roadmap that is there but nobody sees it. So I tend to prefer good, old and chaotic _Issues_ for the sake of simplicity for the community ; )

PR #228, opened 2017-08-01
Refactor tests

jtemporal:
As mentioned in #225 [in a comment](https://github.com/datasciencebr/jarbas/pull/225#issuecomment-318762118) There is a lot of duplicated code that could be shared across test suites.

This is first steps towards removing duplicated code.

**A lot** needs to be done here and I'm not quite sure this is the best way of doing this. Hoping for some feedback here.

### DIFF
--- a/jarbas/core/tests/shared_tests.py
+++ b/jarbas/core/tests/shared_tests.py
@@ -1,3 +1,5 @@
+from unittest.mock import call
+
 from django.test import TestCase
 
 # Serializer methods: TestSerializer
--- a/jarbas/core/tests/shared_tests.py
+++ b/jarbas/core/tests/shared_tests.py
@@ -1,0 +1,18 @@
+from django.test import TestCase
+
+# Serializer methods: TestSerializer
+
+
+def test_serializer(TestCase, obj, expected, input):
+    serialized = obj.serialize(input)
+    TestCase.assertEqual(serialized, expected)
+
+
+# Cotum methods:  TestCustomMethods
+
+
+def test_main(TestCase, obj, update, schedule_update, costum_method):
+    costum_method.return_value = (range(21), range(21, 43))
+    obj.main()
+    update.assert_has_calls([call()] * 2)
+    schedule_update.assert_has_calls(call(i) for i in range(42))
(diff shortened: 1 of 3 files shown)


jarbas/core/tests/shared_tests.py
```diff
@@ -0,0 +1,29 @@
...
+from django.test import TestCase
+
+from jarbas.core.models import Reimbursement
+
+# Serializer methods: TestSerializer
+
+
+def test_serializer(TestCase, obj, expected, input):
+    serialized = obj.serialize(input)
+    TestCase.assertEqual(serialized, expected)
```
cuducos: Instead of passing `TestCase` I'd go for a Jarba's `TestCase` every test on Jarbas inherit from:

```python
from django.test import TestCase as DjangoTestCase


class TestCase(DjangoTestCase):

    def test_serializer(self, obj, expected, input):
        serialized = obj.serialize(input)
        self.assertEqual(serialized, expected)
```

jtemporal:
>What do you think?

I like this a lot!

jtemporal:
hey @cuducos how does dea1f26 looks? in the right path to doing what you mentioned in the PR review?

cuducos:
I like it 👍 

 How do you like it?

The only thing — tiny tiny thing that I bet has already crossed your mind — is placing this `class TestCase` in the beginning of the `__init__.py` — due to its importance. The following auxiliar methods/data/constants could become properties of this object while #178 is still pending (afterwards we can get rid of these auxiliar stuff).

cuducos:
@jtemporal Is this a WIP yet? I fixed a typo and merged master to keep the branch updated. Hope you don't mind. IMHO it's good to merge BTW… just clarify the WIP label so I know what to do ; )

jtemporal:
Don't think so @cuducos the WIP label was removed on aug 7th ;)

jtemporal:
@cuducos Now I think only conflict solving is required cc @anaschwendler

cuducos:
> @cuducos Now I think only conflict solving is required cc @anaschwendler

Just checked the conflicts. IMHO there are pros and cons there:

Going with what is in `refactor-tests` contemplates the purpose of this PR, i.e. DRY. However arguably repetitive tests play around with numbers to test batch sizes and that might get lost by using the structure proposed here. IMHO this is not a a big problem — specially because we tend to get rid of batches and work with async tasks (as in #240). I'd go with changes proposed on `refactor-tests`.

PR #236, opened 2017-08-12
Update docker container versions

Irio:
Wasn't able to fully test it without the changes in the Dockerfile's.

- [x] Update Docker images.
- [x] Update Docker Compose file version.
- [x] Install dependencies I need to properly run `Dockerfile-elm`.


Dockerfile-elm
```diff
@@ -1,13 +1,14 @@
-FROM node:6.9.1
+FROM node:8.3.0-slim
 
-RUN npm install yarn
+RUN npm install gulp yarn -D
```
cuducos: Is `-D` still a shortcut for `--save-dev`? If so this might actually change a file in the project root — which might not be expected.

Also related to that error you've asked in PVT, something like `gulp` not found, installing yarn and gulp as global might be an option.

Issue #211, opened 2017-06-27
Use rows library to convert values


turicas:
jarbas could use some library to automatically identify and convert data types (from strings) instead of doing it itself (like on this [serialize](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/companies.py#L66) method). The [rows library](http://turicas.info/rows) does it automatically and will probably remove a lot of jarbas code if used here, so the import management commands will be easier to write and maintain.

This issue strongly related to https://github.com/datasciencebr/serenata-toolbox/issues/101

cc @cuducos

cuducos:
I think it's a great idea to use `rows` to better handle data input here in Jarbas!

cuducos:
_[Just pasting From (closed) #198]_

[`rows`](http://turicas.info/rows/) already deals with a lot of common issues related to reading a CSV file and serializing data. With that in mind maybe we ca reduce the LOC to maintain in this project if we use `rows` to deal with:

- [ ] [importing and serializing reimbursement data](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/reimbursements.py#L35-L82) (basically handling [this transformation](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/tests/test_reimbursements_command.py#L20-L89))
- [ ] [importing and serializing suspicions data](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/irregularities.py#L37-L71) (basically handling [this transformation](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/tests/test_irregularities_command.py#L18-L35))
- [ ] [importing and serializing companies data](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/companies.py#L34-L77) (basically handling [this transformation](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/tests/test_companies_command.py#L20-L42) and paying attention to the creation of `Activity` data)

Ilozuluchris:
@cuducos,I think I can do this.

cuducos:
Sounds great @Ilozuluchris — I got started one of these days (check #242), so you can either:

* Cover the scripts I haven't covered yet (if I remember right, suspicions and company importion)
* Continue from where I left opening a PR to my branch
* Throw away my code and do it all by yourself

[...]

PR #242, opened 2017-09-02
Use rows to serialize reimbursements

cuducos:
This PR is part of the inclusion of `rows` as suggested in #211, starting with the process of importing reimbursements to Postgres.

Not sure if it's the best implementation so labeled it as an _work in progress_ and I'd like to invite @turicas for a code review. In general my doubts are:

Is there a cleaver way to force conversion of data such as `'2011.0'` to `int`?

https://github.com/datasciencebr/jarbas/blob/7d030103f5c67dea214454173f22f16688a28497/jarbas/core/fields.py#L7-L16

Is there a cleaver way to force conversion of data such as `'2014-02-16T00:00:00'` to `date`?

https://github.com/datasciencebr/jarbas/blob/7d030103f5c67dea214454173f22f16688a28497/jarbas/core/fields.py#L19-L29

Can `rows` rename columns? Actually I'm converting each row (`namedtuple`) to a `OrderedDict`, then to a `dict` and them using this method:

https://github.com/datasciencebr/jarbas/blob/7d030103f5c67dea214454173f22f16688a28497/jarbas/core/management/commands/reimbursements.py#L53-L63

---

Thanks @priscilaportela for pairing!


turicas:
@cuducos, answering the questions:

1- If this is supported right now, then the type identification algorithm will not know when it's an integer or float, leading to some problems. So, by now your way is the best one but I'd replace `value = float(value)` with `value = int(float(value))` since it's expected an `int` object from this `deserialize` method.

2- Yes! You can force the `INPUT_FORMAT` attribute or even create a new class (I prefer the second option):

```python
import rows
# First method
old_format = rows.fields.DateField.INPUT_FORMAT
rows.fields.DateField.INPUT_FORMAT = '%Y-%m-%dT%H:%M:%S'
value = rows.fields.DateField.deserialize('2014-02-16T00:00:00')
rows.fields.DateField.INPUT_FORMAT = old_format

#Second method
class MyDateField(rows.fields.DateField):
    INPUT_FORMAT = '%Y-%m-%dT%H:%M:%S'
value = MyDateField.deserialize('2014-02-16T00:00:00')
```

3- The current way of doing this is using `rows.transform` (you can see [an example here](https://github.com/turicas/rows/blob/develop/examples/library/brazilian_cities_wikipedia.py#L44)) but I'm not sure this is the best solution for these cases (still need to find a better API). The use case of renaming/removing fields is very common, so I'm thinking in a more straightforward way of implementing this on the library. This question is also related to this issue: https://github.com/turicas/rows/issues/30 (it may be a better way of having the conversion you want, but it's not "officially" supported by the library).

jarbas/core/management/commands/reimbursements.py
```diff
@@ -31,53 +34,31 @@ def handle(self, *args, **options):
...
-        missing = ('probability', 'suspicions')
-        for key in missing:
-            reimbursement[key] = None
-
+        force_types = {
+            'cnpj_cpf': TextField,
+            'document_number': TextField,
+            'leg_of_the_trip': TextField,
+            'congressperson_id': IntegerField,
+            'congressperson_document': IntegerField,
```
turicas: If the document starts with 0 you'll loose this information after converting to `int` (`int('00001')` = `1`), so you may want a `TextField` here (same for other document fields).

jarbas/core/fields.py
```diff
@@ -0,0 +1,29 @@
...
+    def deserialize(cls, value, *args, **kwargs):
+        try:
+            # Rows cannot convert values such as '2011.0' to integer
+            value = float(value)
+        except:
+            pass
+        return super(IntegerField, cls).deserialize(value)
+
+
+class DateField(DateField):
```
turicas: You just need to set `INPUT_FORMAT`, don't need to reimplement the method:

```python
class DateField(rows.fields.DateField):
    INPUT_FORMAT = '%Y-%m-%dT%H:%M:%S'
```

jarbas/core/fields.py
```diff
@@ -0,0 +1,29 @@
+from datetime import datetime
+
+from rows.fields import DateField
+from rows.fields import IntegerField as RowsIntegerField
```
turicas: I'd rather use `import rows.fields` and then inherit from `rows.fields.IntegerField`.

jarbas/core/fields.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class IntegerField(RowsIntegerField):
+
+    @classmethod
+    def deserialize(cls, value, *args, **kwargs):
+        try:
+            # Rows cannot convert values such as '2011.0' to integer
+            value = float(value)
```
turicas: Replace with `value = int(float(value))` so the method will always return an `int` object (or `None`).

jarbas/core/management/commands/reimbursements.py
```diff
@@ -31,53 +34,31 @@ def handle(self, *args, **options):
...
+            'reimbursement_value_total': FloatField,
+            'issue_date': DateField,
+            'term': IntegerField,
+            'term_id': IntegerField
+        }
+        with lzma.open(self.path) as file_handler:
+            for row in import_from_csv(file_handler, force_types=force_types):
+                as_dict = dict(row._asdict())  # _asdict returns OrderedDict
+                yield self.rename(as_dict)
```
turicas: You can use `yield from` and centralize all transformation in one method (instead of having a separate method to rename only - this way you can do other transformations as well (if needed) and use this API as a standard to convert data rows). I'd refactor this to:

```python
        with lzma.open(self.path) as file_handler:
            yield from map(self.transform_row,
                           import_from_csv(file_handler,
                                           force_types=force_types))

    @staticmethod
    def transform_row(row):
        """Read the dict generated by rows and fix some keys"""
        rename = {
                'subquota_number': 'subquota_id',
                'reimbursement_value_total': 'total_reimbursement_value',
        )
        return {rename.get(key, key): getattr(row, key)
                for key, value in rename.items()}
    # and remove `rename` method
```

cuducos:
This weekend I gave this PR another try and the results were quite good when combined with Celery. The process of serializing data is basically done by `rows`, which is pretty fast compared to what we had before (30% faster in my local tests, cc @decko).

The PR includes two new libraries when compared to `master`, so in order to test it one have to temporary change the `docker-compose.yml` to use the `Dockerfile`:

```diff
  tasks:
-    image: datasciencebr/jarbas-backend
+    build:
+      context: .
    environment:
```

After this change just `docker-compose build` and then proceeds to the regular tests (in this case, importing reimbursements). If this is a merge, just revert the temporary changes to `docker-compose.yml` ; )

jarbas/core/management/commands/reimbursements.py
cuducos: That makes sense but as those fields are integer in the database (we checked, no leading zeros so far) I decided to define `rows` fields coherent to what the model expects…

jarbas/core/management/commands/reimbursements.py
cuducos: As we're using Celery to process it now, I just converted the `row` to a dictionary and passed it to be processed asynchronously. However I adopted the `transform_row` semantics in `tasks.py` ; ) Tks!

anaschwendler:
Hi @cuducos @turicas !

Nice that we will have rows in our project, great project, and the collaboration is amazing. I'll test that PR my way, but already 👍  !

1. Clone the project:
```console
$ git clone git@github.com:datasciencebr/jarbas.git
```

2. Change to Jarbas folder:
```console
$ cd jarbas
```

3. Checked out to @cuducos PR:
```console
$ git checkout -b cuducos-rows origin/cuducos-rows
```

4. Merge with the master content:
```console
$ git merge master
```

5. Run the project:
```console
$ docker-compose up -d
```

I think that the best way to test if it is working is by running the tests.

5. Run the tests:
```console
$ docker-compose run --rm django python manage.py check
$ docker-compose run --rm django python manage.py test
```

And I got an error:

```
Starting jarbas_elm_1 ... 
Starting jarbas_elm_1 ... done
Starting jarbas_queue_1 ... done
Starting jarbas_tasks_1 ... done
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
....................................................................................................................................................................E
======================================================================
ERROR: jarbas.core.tests.test_reimbursements_command (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: jarbas.core.tests.test_reimbursements_command
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/local/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/code/jarbas/core/tests/test_reimbursements_command.py", line 8, in <module>
    from jarbas.core.management.commands.reimbursements import Command
  File "/code/jarbas/core/management/commands/reimbursements.py", line 4, in <module>
    from rows import import_from_csv
ImportError: No module named 'rows'


----------------------------------------------------------------------
Ran 165 tests in 5.644s

FAILED (errors=1)
Destroying test database for alias 'default'...
````

Did I miss something @cuducos ?

PR #251, opened 2017-09-24
Add link to receipt in the Jarbas Dashboard list view

### LINKED ISSUE #250: Add link to receipt in the Jarbas Dashboard list view
As we have links to Jarbas main view of the receipt, and for Rosie's tweet, we could add an icon so users can jump to the list view (eg home of Jarbas Dashboard) directly to a given receipt PDF.

giovanisleite:
Add link to receipt in the Jarbas Dashboard list view - close #250 

Add a column in dashboard with a link (:page_with_curl:) to the receipt image

Open the link at `dashboard/core/reimbursement/`, will have a :page_with_curl: in between "Reembolso" and "Nome do Parlamentar" in some lines. Filter to the reimbursements of 2012 to see the blank column when don't have a receipt image link.

@cuducos or @jtemporal 

**TODO**

- [x] Add link
- [ ] Automated tests

### DIFF
--- a/jarbas/dashboard/admin.py
+++ b/jarbas/dashboard/admin.py
@@ -237,6 +237,7 @@ class ReimbursementModelAdmin(SimpleHistoryAdmin):
         'short_document_id',
         'jarbas',
         'rosies_tweet',
+        'receipt_link',
         'congressperson_name',
         'year',
         'subquota_translated',
@@ -304,6 +305,19 @@ class ReimbursementModelAdmin(SimpleHistoryAdmin):
 
     rosies_tweet.short_description = ''
     rosies_tweet.allow_tags = True
+
+    def receipt_link(self, obj):
+        try:
+            receipt_url = obj.get_receipt_url()
+            if receipt_url:
+                return '<a target="_blank" href="{}">📃</a>'.format(receipt_url)
+            else:
+                return ''
+        except Exception:
+            return ''
+
+    receipt_link.short_description = ''
+    receipt_link.allow_tags = True
 
     def suspicious(self, obj):
         return obj.suspicions is not None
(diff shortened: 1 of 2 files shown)


jarbas/dashboard/admin.py
```diff
@@ -305,6 +306,19 @@ def rosies_tweet(self, obj):
...
     rosies_tweet.allow_tags = True
 
+    def receipt_link(self, obj):
+        try:
+            receipt_url = obj.get_receipt_url()
+            if receipt_url:
+                return '<a target="_blank" href="{}">📃</a>'.format(receipt_url)
+            else:
+                return ''
+        except Exception:
```
cuducos: What possible exception are we treating here? Generally catching `Exception` is not a best practice IMHO.

But the main point is we can assume [`obj.get_receipt_url()` will return either `None` or a string](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/models.py#L92-L108), specially because [we have already fetched all possible receipt URLs](https://github.com/datasciencebr/jarbas/blob/master/jarbas/core/management/commands/receipts.py).

Yet if we haven't we don't wanna serialize dozens of requests to the Chamber of Deputies servers per user accessing the dashboard. By our side this will response time probably causing a timeout. By the Chamber of Deputies side they might briefly stop responding to requests from our server. So basically the `.get_receipt_url()` is better used asynchronously, outside the HTTP request/response cycle.

So we could simplify your `receipt_link` method as:

```python
def receipt_link(self, obj):
    if not obj.receipt_url:
        return ''
    return '<a target="_blank" href="{}">📃</a>'.format(obj.receipt_url)
```

What do you think about it?

giovanisleite: Thats nice!! I associated force=false with not overloading (only problem that I thought that I could cause) and that is true, so... why use the method? hahahah my bad
 I reviewed and understood my mistake.
Thank you, @cuducos

cuducos: C'mon… it's not a mistake ; ) AFAIK you're new to Jarbas, so it's just a matter of context.

Also you pointed a possible ambiguity in the `force=False` argument: it's only used **if** there is already a `receipt_url` fetched. Otherwise it will start a HTTP request to try to fetch it.

PR #254, opened 2017-09-28
Receipt filter

matheushf:
New filter for receipt


jarbas/core/querysets.py
```diff
@@ -29,6 +29,9 @@ def list_distinct(self, field, order_by_field, query=None):
         self = self.values(field, order_by_field).order_by(order_by_field)
         return self.distinct()
 
+    def receipt(self):
```
cuducos: What about `has_receipt_url` to be clearer? There is so much info about receipts one can get confused (fetched? url? text?)…

jarbas/dashboard/admin.py
```diff
@@ -86,6 +86,19 @@ def queryset(self, request, queryset):
         return queryset.suspicions() if self.value() == 'yes' else queryset
 
 
+class HaveReceiptFilter(JarbasListFilter):
+
+    title = 'existência de recibo'
```
cuducos: What about _Nota fiscal digitalizada_?

jarbas/dashboard/admin.py
```diff
@@ -86,6 +86,19 @@ def queryset(self, request, queryset):
...
+    parameter_name = 'have_receipt'
```
cuducos: `has_receipt` would be better ; )

jarbas/dashboard/admin.py
```diff
@@ -86,6 +86,19 @@ def queryset(self, request, queryset):
         return queryset.suspicions() if self.value() == 'yes' else queryset
 
 
+class HaveReceiptFilter(JarbasListFilter):
```
cuducos: `HasReceipt` would be better ; )

jarbas/dashboard/admin.py
```diff
@@ -311,6 +325,12 @@ def suspicious(self, obj):
     suspicious.short_description = 'suspeito'
     suspicious.boolean = True
 
+    def receipt(self, obj):
```
cuducos: As before: what about `has_receipt_url`?

jarbas/dashboard/admin.py
```diff
@@ -311,6 +325,12 @@ def suspicious(self, obj):
...
+        return obj.receipt_fetched is not False
```
cuducos: This is not the best way to do it. Jarbas could have tried to fetch the receipt (`receipt_fetched = True`) but still the URL wasn't available (`receipt_url = None`). I would put `obj.receipt_url is not False` or `bool(obj.receipt_url)`…

jarbas/core/querysets.py
```diff
@@ -29,6 +29,9 @@ def list_distinct(self, field, order_by_field, query=None):
...
+        return self.exclude(receipt_fetched=False)
```
cuducos: This is not the best way to do it. Jarbas could have tried to fetch the receipt (receipt_fetched = True) but still the URL wasn't available (receipt_url = None). I would use `self.exclude(receipt_url=None)`…

matheushf:
Thanks for the review and the tips @cuducos! I'm getting used with Jarbas and Serenata De Amor. When I get home I'll make these changes.

matheushf:
I believe everything is fine now @cuducos

cuducos:
Hi @matheushf,

I was testing in the browser and I might have found a bug. If I checked _No_ in the _has receipt_ filter you wrote I was still getting reimbursements with receipt.

This is the screen that was supposed to show only reimbursements **with** `receipt_url` (note the highlighted filter):

unknown

On the other hand this is the screen that was supposed to show only reimbursements **without** `receipt_url` (again note the highlighted filter):

unknown

Unfortunately these screens look exactly the same: we have the same reimbursements in both, and a lot of receipt icons in the _No_ screen. Can you check that? I see that my PostgresSQL is really struggling with the large database (I have ~1.6MM reimbursements here) and this might be a bug in my environment.

---

Also I noticed we have no test for the `has_receipt_url` QuerySet, so I tried one in `test_reimbursement_model.py`:

```python
    def test_has_receipt_url(self):
        # let's create a reimbursement with some receipt_url (self.data has none)
        data = self.data.copy()
        data['document_id'] = 42 * 2
        data['receipt_url'] = 'http://serenatadeamor.org/'
    
        # now let's save two reimbursements: one with and another one without receipt_url
        Reimbursement.objects.create(**data)

        self.assertEqual(1, Reimbursement.objects.suspicions().count())
```

Would you mind adding it to the PR, or if don't fancy it opening an issue so we remember to write it one of these days?

giovanisleite:
@cuducos @matheushf On your environment, when you filter the yes and no for suspicious reimbursements, the filter works? Because when I try on mine the 'yes' works, but the no doesn't. Yet, I  in production works :fearful:

This part looks wrong to me, is it?

`#admin.py`
`return queryset.suspicions() if self.value() == 'yes' else queryset`

When is chose 'yes' on the filter, the set is filtered. But when the 'no' is chosen, the result is equal to 'all'? (is that what happens on dev environment)

matheushf:
Hello @cuducos and @giovanisleite,

I've noticed that it's being really slow for PostgresSQL to realize these queries, in both suspicions and has_receipt_url, because when I add more filters to the query (so it will bring me just a few results) it works pretty fast.  

I've made some tests here, and I don't know why it isn't working for you guys. I've manually added a url_receipt in the database, to manually test the filter, and as showed in this screenshot, it's working fine. 

captura de tela de 2017-09-30 23-20-13
captura de tela de 2017-09-30 23-20-34

I have also added the unit test you wrote, and it worked fine (after I changed from `Reimbursement.objects.suspicions()` to `Reimbursement.objects.has_receipt_url()` )

captura de tela de 2017-10-01 16-29-43

giovanisleite:
Now it is working like the suspicions one. :tada:

I am a litte bit confuse about the function of the 'no' of these filters, so I will open another issue about that.

cuducos:
> Now it is working like the suspicions one. 🎉

Thus according to #256 recently open it has the same bug as the suspicions filter haha…

Would you mind fixing the bug for the receipt URL filter so we can merge this?

matheushf:
Great @giovanisleite!
I fixed it for receipt_url @cuducos, I believe it's ok to merge now. :)

jarbas/api/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
         if self._bool_param('suspicions'):
             self.queryset = self.queryset.suspicions()
 
+        # filter has_receipt_url
+        has_receipt_url = self._bool_param('has_receipt_url')
+        if has_receipt_url is not None:
+            self.queryset = self.queryset.has_receipt_url(has_receipt_url)
+
```
cuducos: I was thinking… If we're offering this filter for the JSON API we need to document that in the `README.md` and write tests for that. You can either update the documentation and add the tests, or stash this change for another PR focusing on the API.

jarbas/dashboard/admin.py
```diff
@@ -88,6 +88,19 @@ def queryset(self, request, queryset):
...
+
+    title = 'nota fiscal digitalizada'
+    parameter_name = 'has_receipt'
+    options = (
+        ('yes', 'Sim'),
+        ('no', 'Não'),
+    )
+
+    def queryset(self, request, queryset):
+        return queryset.has_receipt_url(self.value() == 'yes') if self.value() else queryset
```
cuducos: Shouldn't this be something following this logic?

* If filter is `yes` we show only reimbursements **with** `receipt_url`, i.e. `self.queryset.has_receipt(True)`
* If filter is `no` we show only reimbursements **withot** `receipt_url`, i.e. `self.queryset.has_receipt(False)`
* If there's no filter (or something else than `yes` or `no`) we show everything, i.s. `self.queryset`

It's a three possibilities switch but your code offers just two. [This comment elsewhere](https://github.com/datasciencebr/jarbas/pull/257#discussion_r142478743) might be useful ; )

jarbas/api/views.py
matheushf: I think I'll add it to this PR, since it's quite a small change. What you think?

jarbas/dashboard/admin.py
matheushf: This is indeed a better way!

README.md
```diff
@@ -62,6 +62,7 @@ All these endpoints accepts any combination of the following parameters:
 * `month`
 * `subquota_id`
 * `suspicions` (_boolean_, `1` parses to `True`, `0` to `False`)
+* `has_receipt` (_boolean_, `yes` parses to `True`, `no` to `False`)
```
matheushf: It's just about this line, right?

jarbas/dashboard/admin.py
```diff
@@ -98,7 +98,11 @@ class HasReceiptFilter(JarbasListFilter):
     )
 
     def queryset(self, request, queryset):
-        return queryset.has_receipt_url(self.value() == 'yes') if self.value() else queryset
+        filter = {
+            'yes': queryset.has_receipt_url(True),
+            'no': queryset.has_receipt_url(False)
+        }
+        return filter.get(self.value(), queryset)
```
cuducos: My bad here. Using `filter` is not recommended (because it overrides the native `filter` function). Choose anything else, even `filter_` instead… so sorry, my mistake not yours ; )

README.md
cuducos: Yes… but… may we keep the sabe _bool string_ documented for both suspicions and receipt? For suspicions we say `1` our `0`, for receipt you suggested `yes` or `no`…

jarbas/api/views.py
cuducos: That'd be great ; ) Tks!

README.md
matheushf: I forgot to ask that before comiting, but, since you are using 'yes' and 'no', why here it's 1 and 0?

jarbas/dashboard/admin.py
matheushf: Actually I thought about it, but again, forgot to ask. Can I get in contact in the telegram group now for things like this?

README.md
cuducos: Addressing your question directly: the pint is _where_ we're using `1`/`0` and _where_ we're using `yes`/`no`. For the [dashboard (`admin.py`)](https://jarbas.serenatadeamor.org/dashboard/core/reimbursement) we use `yes`/`no`. For the [JSON API](https://jarbas.serenatadeamor.org/api/reimbursement/) we use `1`/`0`.

Why? Basically _bad design_, uninformed decisions in the past that _totally_ should be fixed — but I think this is an issue for a new PR ; )

jarbas/dashboard/admin.py
cuducos: Sure thing ; )

jarbas/api/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
         if self._bool_param('suspicions'):
             self.queryset = self.queryset.suspicions()
 
+        # filter has_receipt_url
+        has_receipt_url = self._bool_param('has_receipt_url')
```
cuducos: You documented `has_receipt`, but here you used `has_receipt_url`… can you fixed that, please? I think we can change it here (and have `receipt_url` instead of `has_receipt_url` as the variable name).

If write a test for this view with the filter would be great too ; )

jarbas/core/tests/test_reimbursement_model.py
```diff
@@ -89,6 +89,29 @@ def test_suspicions(self):
         Reimbursement.objects.create(**self.data)
         self.assertEqual(1, Reimbursement.objects.suspicions().count())
 
+    def test_has_receipt_url(self):
```
matheushf: Do you mean this tests @cuducos?

jarbas/api/views.py
matheushf: I wrote filters for has_receipt and has_no_receipt, commented inline, if it's what you meant. :)

jarbas/core/tests/test_reimbursement_model.py
giovanisleite: He is talking about the line 65 of readme, @matheushf. The param there is "has_receipt", but on view you call for "has_receipt_url" (line 34)

jarbas/api/views.py
```diff
@@ -30,6 +30,11 @@ def get(self, request):
         if self._bool_param('suspicions'):
             self.queryset = self.queryset.suspicions()
 
+        # filter receipt_url
+        receipt_url = self._bool_param('receipt_url')
```
giovanisleite: Hi @matheushf, sorry for my last message, i talked in another place and I feel that I confused you.
README tells that this param is called 'has_receipt', but here you call him of 'receipt_url'. So, pick one :smile: .
I think receipt_url is more like suspicions style, without verb. But has_receipt tells more. So.. i don't know hahaha

cuducos:
Yay, many thanks, @matheushf — awesome contribution ; )

PR #252, opened 2017-09-27
Use full text search to query reimbursements

cuducos:
9df9666 starts this WIP PR targeting #226

**What is already done**

* Adds `search_vector` field to `Reimbursement` model (with `None` as default value)
* Creates script to fill in the `search_vector` field (takes minutes/hours)

**How to test it**

1. First apply the new migrations<br>`$ python manage.py migrate`
1. Run the new search vector command<br>`$ python manage.py searchvector`
1. Play with the Django console:<br>`$ python manage.py shell_plus`

For example:

```python
from django.db.models import F
from django.contrib.postgres.search import SearchVector, SearchQuery, SearchRank

query = SearchQuery('buffet', config="portuguese")  # create a full text search query
rank = SearchRank(F('search_vector'), query)  # create a ranking based on this query

Reimbursement.objects.annotate(rank=rank) \
    .filter(search_vector=query) \
    .order_by('-rank')  # you should get a list of reimbursements with the world 'buffet'
```

**Next steps**

Adapt the dashboard search box to use this feature.

PR #255, opened 2017-09-29
Fix bug when using full text search w/ filters

cuducos:
#252 introduced a small but nasty bug: in the dashboard if you use the search box all filters are ignored. This PR fixes this bug.

How to test it: open the dashboard, select a filter (for example, list only suspicions reimbursements) and then try a text search from the search box. If the results show the filter is combined with the text search it worked, yay!

### DIFF
--- a/jarbas/dashboard/admin.py
+++ b/jarbas/dashboard/admin.py
@@ -370,17 +370,17 @@ class ReimbursementModelAdmin(SimpleHistoryAdmin):
         return super().formfield_for_dbfield(db_field, **kwargs)
 
     def get_search_results(self, request, queryset, search_term):
-        if not search_term:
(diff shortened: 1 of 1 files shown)


anaschwendler:
What I did to test this PR

1 - Clone the project
```console
$ git clone git@github.com:datasciencebr/jarbas.git
```

2 - Get in the folder
```console
$ cd jarbas
```

3 - Checkout to @cuducos PR
```console
$ git checkout -b cuducos-combine-full-text-search-ans-filters origin/cuducos-combine-full-text-search-ans-filters
```

4 - Run the enviroment
```console
$ make run.devel
```

4 - Access `localhost:8000`, and the result:
With suspicions only filter:
image

Searching for `restaurante`:
image

PR #257, opened 2017-10-02
Fix suspicions filter

### LINKED ISSUE #256: The 'yes'/'no' dashboard filters
Hi,

I was looking the pull request #254 (implementing the has_receipt_url filter that is like the suspicions), and now I am confuse about the operation of the filter.

How should these filters work?
e.g. "suspicions"
"All": shows all the reimbursements;
"Yes": Show only reimbursements that are suspect;
"No": Show only reimbursements that aren't suspect. Right?

So... why there isn't a query for the 'no'? Like `
return self.exclude(suspicions__isnull=True)
`?
for the 'no', returns just the queryset that is the same return for 'all' (ausence of value).

If I am right about this, there is a bug on the 'no' filter

Like [this](https://jarbas.serenatadeamor.org/dashboard/core/reimbursement/?month=12&state=SP&subquota_id=13&year=2016), using another filters just too see the number of results.
Filters that I used: Month: December, Year: 2016. State: SP. Subquota: 'Congressperson meal' (id 13).

Without suspicious filter (selected: 'All'): 259 results.
When I try 'yes' for suspicious, 1 result.
When I try 'no' for suspicious, 259 results. I think that isn't working, because should be 258 results (1 suspect + 258 that isn't suspect).

Am I missing something?

giovanisleite:
Fix the 'no' option for suspicions filter - fix #256  

Add a boolean as param to the suspicions queryset. The boolean is true when the value of the filter is 'yes', so it returns just the suspects and false when the value of the filter is 'no', so this method returns just the reimbursements that are not a supect.

Open dashboard, click on Reimbursements, click on 'no' for the suspicions filter. if it works, the list of reimbursements shouldn't show the suspects reimbursements (Those that the last column is true) .

@cuducos or @jtemporal 

I am not sure about the tests that I wrote, so.. special attention there.
Thank you.

### DIFF
--- a/jarbas/api/views.py
+++ b/jarbas/api/views.py
@@ -27,8 +27,9 @@ class ReimbursementListView(ListAPIView):
         filters = {k: v for k, v in zip(params, values) if v}
 
         # filter suspicions
-        if self._bool_param('suspicions'):
-            self.queryset = self.queryset.suspicions()
+        suspicions = self._bool_param('suspicions')
+        if suspicions is not None:
+            self.queryset = self.queryset.suspicions(suspicions)
 
         # filter reimbursement in latest dataset
         in_latest = self._bool_param('in_latest_dataset')
--- a/jarbas/dashboard/admin.py
+++ b/jarbas/dashboard/admin.py
@@ -85,7 +85,7 @@ class SuspiciousListFilter(JarbasListFilter):
     )
 
     def queryset(self, request, queryset):
-        return queryset.suspicions() if self.value() == 'yes' else queryset
+        return queryset.suspicions(self.value() == 'yes') if self.value() else queryset
 
 
 class MonthListFilter(JarbasListFilter):
--- a/jarbas/dashboard/admin.py
+++ b/jarbas/dashboard/admin.py
@@ -85,7 +85,11 @@ class SuspiciousListFilter(JarbasListFilter):
     )
 
     def queryset(self, request, queryset):
-        return queryset.suspicions(self.value() == 'yes') if self.value() else queryset
+        filter_option = {
+            'yes': queryset.suspicions(True),
+            'no': queryset.suspicions(False)
+        }
+        return filter_option.get(self.value(), queryset)
 
 
 class MonthListFilter(JarbasListFilter):
--- a/jarbas/core/querysets.py
+++ b/jarbas/core/querysets.py
@@ -29,8 +29,8 @@ class ReimbursementQuerySet(models.QuerySet):
         self = self.values(field, order_by_field).order_by(order_by_field)
         return self.distinct()
 
-    def suspicions(self):
-        return self.exclude(suspicions=None)
+    def suspicions(self, boolean):
+        return self.exclude(suspicions=None) if boolean else self.filter(suspicions=None)
 
     def in_latest_dataset(self, boolean):
         return self.filter(available_in_latest_dataset=boolean)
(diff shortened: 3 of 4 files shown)


jarbas/dashboard/admin.py
```diff
@@ -85,7 +85,7 @@ class SuspiciousListFilter(JarbasListFilter):
     )
 
     def queryset(self, request, queryset):
-        return queryset.suspicions() if self.value() == 'yes' else queryset
+        return queryset.suspicions(self.value() == 'yes') if self.value() else queryset
```
cuducos: I think here we have a three state choice, don't we? Specially now we're paying attention to the so far unknown bug…

* If the filter is `yes` users should see only reimbursements **with suspicions**, i.e. `queryset.suspicions(True)`
* If the filter is `no`users should see only reimbursements **withot suspicions**, i.e. `queryset.suspicions(False)`
* If filter is not selected, the users see everything, i.e. `queryset`

Does that make sense?

giovanisleite: Yes, I see that way.
Do you think that organize the code like that could help to understand?

cuducos: Sure thing… I'd draft something among these lines:

```python
def queryset(self, request, queryset):
    filter = {
        'yes': queryset.suspicions(True),
        'no': queryset.suspicions(False)
    }
    return filter.get(self.value(), queryset)
```

cuducos:
In addition to that, do you fancy fixing the JSON API too?

Probably a simple fix would be this strategy @matheushf made [in another PR](https://github.com/matheushf/jarbas/blob/62fc68b1e8700078c49261634932574e7a40d60e/jarbas/api/views.py#L33-L36) ; )

I'm happy to see this change here, but feel free to open another PR, or just dismiss my suggestion… just LMK and I'll add another issue so we don't forget.

jarbas/api/views.py
```diff
@@ -27,8 +27,9 @@ def get(self, request):
         filters = {k: v for k, v in zip(params, values) if v}
 
         # filter suspicions
-        if self._bool_param('suspicions'):
-            self.queryset = self.queryset.suspicions()
+        suspicions = self._bool_param('suspicions')
```
giovanisleite: @cuducos, For the API, is it?

jarbas/api/views.py
```diff
@@ -27,8 +27,9 @@ def get(self, request):
...
+        if suspicions is not None:
+            self.queryset = self.queryset.suspicions(suspicions)
```
giovanisleite: @cuducos, for the API, is it? I already did this part.
If there is more, let me know.

jarbas/dashboard/admin.py
giovanisleite: This is a nice way to handle the three choices, thanks @cuducos.

jarbas/api/views.py
cuducos: Damn… my bad! I missed it. Sorry, mate.

jarbas/dashboard/admin.py
cuducos: Just beware that using `filter` is not recommended (because it overrides the native `filter` function). Choose anything else, even `filter_` instead ; )

jarbas/api/views.py
giovanisleite: two identical PR hahahah Yesterday I was looking to one of them.. "But I saw some comment about (insert here some part), where is it?? 😖"
I will fix the rest today, many thanks, @cuducos

giovanisleite:
Done, @cuducos :smiley:

cuducos:
Yay… many thanks, @giovanisleite — great contribution! Deploying later today.

PR #262, opened 2017-10-04
Importing and serializing company data

Ilozuluchris:
Quick check,please am I on the right path #211

### DIFF
--- a/jarbas/core/management/commands/companies.py
+++ b/jarbas/core/management/commands/companies.py
@@ -1,6 +1,7 @@
 import csv
 import lzma
-
+from rows.fields import EmailField,DatetimeField,FloatField
+import rows
 from django.core.exceptions import ValidationError
 from django.core.validators import validate_email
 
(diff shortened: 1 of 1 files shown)


jarbas/core/management/commands/companies.py
```diff
@@ -1,6 +1,6 @@
 import csv
 import lzma
-
```
cuducos: This line separating native modules imports and third-party module imports should be kept (as it's a PEP8 recommendation)

jarbas/core/management/commands/companies.py
```diff
@@ -1,6 +1,6 @@
...
+from rows.fields import EmailField,DatetimeField,FloatField
```
cuducos: Would you mind using a space after the commas (PEP8 recommendation)?

cuducos:
I added minor comments on code style. You can use [`prospector`](https://prospector.landscape.io/) to assist you:

```console
$ pip install prospector
$ prospector -s veryhigh --max-line-length 120 jarbas
```

You can replace `jarbas` by the path of the files/modules you edited to see the report of your changes only ; )

Ilozuluchris:
@cuducos Thanks I'll do that

PR #263, opened 2017-10-07
Importing and serializing company data

Ilozuluchris:
This is a WIP but I want to invite @cuducos  to review what I have done so far #211


jarbas/core/management/commands/companies.py
```diff
@@ -1,5 +1,6 @@
-import csv
 import lzma
+import rows
+import rows.fields
```
cuducos: Do you really need to import `rows` and then `rows.fields`?

jarbas/core/management/commands/companies.py
```diff
@@ -23,6 +24,8 @@ def handle(self, *args, **options):
 
         self.save_companies()
 
+
+
```
cuducos: We don't need these extra spaces: according to PEP8 1 line is fine to separate class methods.

jarbas/core/management/commands/companies.py
```diff
@@ -45,6 +48,12 @@ def save_companies(self):
                 self.count += 1
                 self.print_count(Company, count=self.count)
 
+    @staticmethod
+    def transform(row):
+        return row._asdict()
+
+
+
```
cuducos: We don't need these extra spaces: according to PEP8 1 line is fine to separate class methods.

jarbas/core/management/commands/companies.py
```diff
@@ -84,3 +91,7 @@ def to_email(email):
 
         except ValidationError:
             return None
+
+
+class InputDateField(rows.fields.DateField):
+    INPUT_FORMAT = '%d/%m/%Y'
```
cuducos: I'd declare this class in the beginning of the file so when you use it in the other class the reader is aware about what this field is ; )

jarbas/core/management/commands/companies.py
```diff
@@ -45,6 +48,12 @@ def save_companies(self):
                 self.count += 1
                 self.print_count(Company, count=self.count)
 
+    @staticmethod
+    def transform(row):
+        return row._asdict()
```
cuducos: I don't think we need a method for that… calling `_asdict()` in the `for` loop seams enough.

jarbas/core/management/commands/companies.py
```diff
@@ -63,17 +72,15 @@ def save_activities(self, row):
 
         return [main], secondaries
 
+
     def serialize(self, row):
```
cuducos: Basically the great point of using `rows` is to get rid of the `serialize` method. The idea is to use the `force_types` keyword argument when calling `rows.import_from_csv` and let `rows` serialize stuff as we need it ([for example](https://github.com/datasciencebr/jarbas/blob/8dfb315ed6affe0026507c8a543c6c3752806cef/jarbas/core/management/commands/reimbursements.py#L37-L49), #211).

cuducos (changes requested):
Hi @Ilozuluchris! One more many thanks for the WIP PR. First of all I changed the merge branch to `master`, otherwise it would be impossible to browser the changes and picking up only what you've actually changed and what was merged since I created `cuducos-rows` branch.

That said you're in the right path. The main point is that `rows` powerful import tool can handle all serialization so we can get rid of that in Jarbas source code. So there is still a few extras steps (mostly related to the `force_types` kwarg).

In addition, there are some Python style issues that are indeed minor stuff. Just run `prospector` before you declare you code ready and probably you can spot them all ; )

jarbas/core/management/commands/companies.py
```diff
@@ -31,7 +34,7 @@ def save_companies(self):
         skip = ('main_activity', 'secondary_activity')
         keys = tuple(f.name for f in Company._meta.fields if f not in skip)
         with lzma.open(self.path, mode='rt', encoding='utf-8') as file_handler:
-            for row in csv.DictReader(file_handler):
+            for row in map(self.transform,rows.import_from_csv(filename_or_fobj=file_handler)):
```
cuducos: This might work but this is not pythonic. First of all, style:

1. [try to keep things under 80 chars](https://www.youtube.com/watch?v=wf-BqAjZb8M)
1. list comprehension _usually_ are more readable than `map` - Zen of Pytthon
1. always add a space after come when separating values (in sequences) and arguments (in function calls) — also PEP8

I'd try something like:

```
for row in rows.import_from_csv(file_handler):
    row = row._asdict()  # and we can get rid of the self.transform call
    …
```

cuducos (changes requested):
I forgot to add that comment on the previous review, sorry.

jarbas/core/management/commands/companies.py
Ilozuluchris: Okay @cuducos Thanks for the review,I would make the necessary changes .

Ilozuluchris (commented):
@cuducos rows.fields.EmailField raises a value error when it tries to serialize 'ahoy'

cuducos:
> `rows.fields.EmailField` raises a value error when it tries to serialize `'ahoy'`

IMHO that is expected as `'ahoy'` is not a valid email address… right?

Ilozuluchris:
Yeah,it is...the `serialize` method returned invalid email addresses as `none`.I just mentioned it since that was not how the serialized handled invalid email addresses.

Ilozuluchris:
@cuducos I am confused on how to handle this https://github.com/datasciencebr/jarbas/blob/f95addd5ea4c47f161796cf1e72a58050f66ce60/jarbas/core/management/commands/suspicions.py#L50   since the serialize method actually reshapes rows

cuducos:
> @cuducos I am confused,I do not know how to go about this

Would you mind expanding on your doubt? I mean… you just highlighted the method we're trying to get rid of, so I don't know exactly where you doubt lies…

Ilozuluchris:
reserved_keys = (
            'applicant_id',
            'document_id',
            'probability',
            'year'
        )
        hypothesis = tuple(k for k in row.keys() if k not in reserved_keys)
        pairs = ((k, v) for k, v in row.items() if k in hypothesis)
        filtered = filter(lambda x: self.bool(x[1]), pairs)
        suspicions = {k: True for k, _ in filtered} or None

        return dict(
            document_id=document_id,
            probability=probability,
            suspicions=suspicions
        )
I was talking about this arrangement,am not sure how to go about this with rows

cuducos:
Ow… you're right! I don't know any thing by heart, maybe here we do need a serialization method!… @turicas, any idea on that?

Our input is a CSV like:

```csv
document_id, hypothesis_1, hypotehesis_2, hypothesis_3
42,True,False,True
```

And the expected output is something like:

```pyhon
[
    {
        'document_id': 42,
        'suspicions': {
            'hypothesis_1': True,
            'hypothesis_3': True
        }
    }
]
```

Ilozuluchris:
Hello @cuducos, anything left wrt issue  #211

cuducos (approved):

cuducos:
> Hello @cuducos, anything left wrt issue #211

Yep, tests are failing, so there is still something to fix. You might consider adding rows to the `requirements.txt`; )

Ilozuluchris:
Okay  I will, thanks.

Ilozuluchris:
@cuducos The travis CI job does not run, the requests page reports this
> GitHub payload is missing a merge commit (mergeable_state: "unknown", merged: false) 

I believe it has to do with the merge conflict wrt requirements.txt. I tried to resolve myself but that did not work as expected.

cuducos:
> The travis CI job does not run

After fixing conflicts it is still red:

```
======================================================================
ERROR: test_save_companies (jarbas.core.tests.test_companies_command.TestCreate)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/python/3.5.4/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/python/3.5.4/lib/python3.5/unittest/case.py", line 605, in run
    testMethod()
  File "/opt/python/3.5.4/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
TypeError: test_save_companies() missing 1 required positional argument: 'lzma'
```

cuducos:
@Ilozuluchris, I'm not sure why the CodeClimate checks are missing. Gonna fix that and then we come back to this PR, ok?

Ilozuluchris:
@cuducos Okay no problem, I am sorry it took so long to do this.

jarbas/core/tests/test_companies_command.py
```diff
@@ -61,22 +34,24 @@ def test_save_activities(self, update_or_create):
...
+                                         'fixtures', 'companies.xz')
         self.command.save_companies()
-        create.assert_called_with(ahoy=42)
+        create.assert_called_with(additional_address_details=b'', address=b'', city=b'', cnpj=b'',
+                                  email='test@test.com.br', last_updated=b'', latitude=-15.7910966, legal_entity=b'',
+                                  longitude=-47.9508743, name='Test', neighborhood=b'', number=1, opening=None,
+                                  phone=b'', responsible_federative_entity=b'', situation=b'', zip_code=b'',
+                                  situation_date=datetime.date(2005, 9, 24), situation_reason=b'', type='Book',
+                                  special_situation_date=None, state=b'', status=b'', trade_name=b'',
+                                  special_situation=b'')
```
cuducos: What about creating a dictionary and unpacking it to enhance the readability here?

```python

kwargs = {
    'additional_address_details': b'',
    'address': b'',
    …
}
create.assert_called_once_with(**kwargs)
```

cuducos (commented):
Hi @Ilozuluchris, many thanks for pushing this PR further. I'm just afraid CodeClimate could help you with code style. Would you mind pushing a new commit so the CodeClimate check runs? (I just fixed it).

jarbas/core/tests/test_companies_command.py
Ilozuluchris: Okay, I will.

Ilozuluchris:
@cuducos  All checks passed.

jarbas/core/management/commands/companies.py
```diff
@@ -1,13 +1,24 @@
...
+
+
+companies_csv_field_types = {
+            'email': rows.fields.EmailField,
+            'opening': CompaniesDate,
+            'situation_date': CompaniesDate,
+            'special_situation_date': CompaniesDate,
+            'latitude': rows.fields.FloatField,
+            'longitude': rows.fields.FloatField
+        }
```
cuducos: This is a contant, could you named it with caps (i.e. `COMPANIES_CSV_FIELDS`) and put it before the `class` definition (just after the imports)?

Also, in Python we use one of these formatting styles:

```python
COMPANIES_CSV_FIELDS = {
    'email': rows.fields.EmailField,
    'opening': CompaniesDate,
    'situation_date': CompaniesDate,
    'special_situation_date': CompaniesDate,
    'latitude': rows.fields.FloatField,
    'longitude': rows.fields.FloatField
}
```

Or:

```python
COMPANIES_CSV_FIELDS = {'email': rows.fields.EmailField,
                        'opening': CompaniesDate,
                        'situation_date': CompaniesDate,
                        'special_situation_date': CompaniesDate,
                        'latitude': rows.fields.FloatField,
                        'longitude': rows.fields.FloatField}
```

Also, can you reflect this formatting decision in `arbas/core/tests/test_companies_command.py` (line 46), isting one field per line?

cuducos (changes requested):
Hi, just some formatting issues (don'tknow why) Code Climate skips.

PR #290, opened 2017-11-28
Fix #289

### LINKED ISSUE #289: The refactored docker image broke celery?
Hey folks,

I am having problems to load the sample data

image

The #285 introduced a new api to docker and seems that all commands are preceded by `python manage.py`, but that was supposed to happen on tasks too?
Looks like tasks are trying to use the command `celery worker --app jarbas` and getting `python manage.py celery worker --app jarbas`.

I was able to use the celery workers/tasks, just running `docker-compose exec -d  tasks celery worker --app jarbas`

Am I doing something wrong?
Thanks.

caduvieira:
New entrypoint for service 'task'

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,8 @@ services:
     depends_on:
        - postgres
        - queue
-    command: [celery, worker, --app, jarbas]
+    entrypoint: ['/bin/sh', '-c']
+    command: ['celery worker --app jarbas']
 
   elm:
     image: datasciencebr/jarbas-frontend


caduvieira:
I returned the '/bin/sh -c' entrypoint

This is only necessary if we are going to run more than one command other than celery.

If only celery is going to be running we could change the entrypoint to celery like 
https://github.com/datasciencebr/jarbas/blob/master/docker-compose.prod.yml#L15

cuducos:
Good strategy — this gives us more flexibility IMHO!

cuducos:
Many many thanks @caduvieira : )

PR #314, opened 2018-02-07
Remove PostgreSQL from Docker architecture for production

cuducos:

Running PostgreSQL in our Digital Ocean droplets was too consuming, we needed to fix that.

We moved the database to a dedicated service for now.

Access [jarbas.serenata.ai](https://jarbas.serenata.ai) as it is already using this new setup (it was deployed with this branch not `master` after last week crash).

@Irio @anaschwendler @jtemporal (who loves Docker when it works, which seams to be the case)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,28 +5,28 @@ services:
     image: datasciencebr/jarbas-backend
     environment:
        - DATABASE_URL=postgres://jarbas:mysecretpassword@postgres/jarbas
-       - ALLOWED_HOSTS=localhost,127.0.0.1
-       - AMAZON_S3_BUCKET=serenata-de-amor-data
-       - AMAZON_S3_REGION=s3-sa-east-1
-       - AMAZON_S3_CEAPTRANSLATION_DATE=2016-08-08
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -4,41 +4,29 @@ services:
...
-       - CELERY_BROKER_URL=amqp://guest:guest@queue/
+      - CELERY_BROKER_URL=amqp://guest:guest@queue/
     depends_on:
-       - postgres
-       - queue
-    entrypoint: ['/bin/sh', '-c']
-    command: ['celery worker --app jarbas']
+      - queue
+    entrypoint: ["/bin/sh", "-c"]
+    command: ["celery worker --app jarbas"]
```
Irio: The idea of having commands as an array is to split it by spaces.

https://docs.docker.com/compose/compose-file/#command
https://docs.docker.com/engine/reference/builder/#cmd

```
command: ["celery", "worker", "--app", "jarbas"]
```

cuducos:
> Good to merge once you update the code as asked.

Yay! Done in b27b407.

